PR #3, opened 2017-05-23
Add initial repo metadata

marcan:


metadata/layout.conf
```diff
@@ -0,0 +1,3 @@
+masters = gentoo
+cache-formats = md5-dict
```
simonvanderveldt: @marcan what do `cache-format` and `thin-manifests` do?

marcan: `cache-formats` specifies the supported cache metadata formats. For a new repo, it makes sense to enable only the current md5-dict format.
`thin-manifests` disables ebuild digest information in manifests. This goes together with Git repos, which already contain their own hashes which make them unnecessary (and allows for merging ebuild changes sanely - otherwise you'd have to recompute the digest any time a merge is done, making version control painful).

simonvanderveldt: @marcan Cool, thanks for the info!

mrVanDalo:
```#eapis-banned = 1
eapis-deprecated = 0 1 2 3 4
```

would be nice to ban old builds (maybe `0 - 2`) and deprecated `3 - 4`

mrVanDalo:
```
manifest-hashes = SHA256 SHA512 WHIRLPOOL
```

would also be nice to use sha for hashes

marcan:
@mrVanDalo maybe eapis-banned to make repoman hard-fail?

simonvanderveldt:
> would also be nice to use sha for hashes

Do we need to define these? AFAIK they are defaulted in portage already
(and since there was some discussion on `#gentoo-portage` about changing this it probably makes sense to just use the default)

mrVanDalo:
> AFAIK they are defaulted in portage already
Yeah true.

marcan:
Let's open a new PR for the EAPIs so this one doesn't drag on forever :)

PR #4, opened 2017-05-23
add media-sound/bitwig-studio : version 1.3.16

mrVanDalo:
I also created this ticket to get a feeling on what is important for use and what not.


simonvanderveldt:
@mrVanDalo Did you deliberately choose the old version? 2.0 and now 2.1 have been out for a bit, would make more sense to package those, right?

These are the properties of the 2.1 `deb`
```
~/Downloads ❯ dpkg -I bitwig-studio-2.1.deb 
 new debian package, version 2.0.
 size 267722934 bytes: control archive=670 bytes.
     656 bytes,    13 lines      control              
     223 bytes,     3 lines      md5sums              
 Package: bitwig-studio
 Provides: bitwig-studio
 Replaces: bitwig-studio
 Version: 2.1
 Architecture: amd64
 Maintainer: Bitwig GmbH <support@bitwig.com>
 Installed-Size: 290712
 Depends: libx11-xcb1, libstdc++6, libc6, libgcc1, libpixman-1-0, libfontconfig1, libx11-6, zlib1g, libxfixes3, libexpat1, libxrender1, libx11-6, zlib1g, libjack-jackd2-0, xdg-utils, libav-tools, zenity
 Section: Multimedia
 Priority: extra
 Homepage: http://bitwig.com
 Description: Music production and performance system
   Bitwig Studio is a multi-platform music-creation system for production, remixing and performance with a focus on flexible editing tools and a super-fast workflow.
```
Also, I think it would be nice to have a demo ebuild as well so people can easily install the demo, what do you think?

And have you tried it with jack1?

mrVanDalo:
bitwig2 would get a new slot. I only have a license for bitwig1. 
The binary you download there is also the demo. When you install it, you can choose.

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,84 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+# TODO
+# - Verify all runtime dependencies, min versions, and required flags
+# - Avoid bundled JRE?
+# - use EAPI=6
+
+EAPI=5
```
mrVanDalo: or irc we agreed on eapi6

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,84 @@
...
+RDEPEND="${DEPEND}
+		app-arch/bzip2
+		dev-libs/expat
+		gnome-extra/zenity
+		media-libs/alsa-lib
+		media-libs/fontconfig
+		media-libs/freetype
+		media-libs/libpng:0/16
+		media-libs/mesa
+		media-sound/jack2
```
mrVanDalo: on irc we agreed on using  virtual/jack

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,84 @@
...
+EAPI=5
+inherit eutils fdo-mime unpacker
+
+DESCRIPTION="for creation and performance of your musical ideas on stage or in the studio."
+HOMEPAGE="http://bitwig.com/"
+SRC_URI="http://downloads.bitwig.com/${P}.deb"
+LICENSE="Bitwig"
+SLOT="1"
+KEYWORDS="~amd64"
+RESTRICT="mirror strip"
```
simonvanderveldt: @mrVanDalo is the strip RESTRICT necessary? I'm assuming the binaries are stripped, but you never know.
Also we could make this more specific by using `QA_PRESTRIPPED`. Would that be better/beneficial?

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,84 @@
...
+
+	insinto /usr/share/mime/packages
+	doins usr/share/mime/packages/bitwig-studio.xml
+}
+
+pkg_postinst() {
+	fdo-mime_mime_database_update
+	fdo-mime_desktop_database_update
+
+	if ! use libav; then
```
simonvanderveldt: @mrVanDalo I'm not sure this is still necessary, see quote below from https://downloads.bitwig.com/stable/1.3.9/Release-Notes-1.3.9.html

> IMPROVED	Search for ffprobe/ffmepg or avprobe/avconf on Linux.

mrVanDalo: your right 
I'll fix it

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,90 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+# TODO
+# - Verify all runtime dependencies, min versions, and required flags
```
simonvanderveldt: I think we can remove this one, right?
With the current DEPENDS it works as it should, don't think we need specific versions at this point, right?

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,82 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+# TODO
+# - What about bundled libs? Shouldn't they be removed?
+# - Install docs in correct location? (currently in /opt/bitwig-studio/resources/doc)
+
+EAPI=6
+inherit eutils fdo-mime unpacker
+
+DESCRIPTION="for creation and performance of your musical ideas on stage or in the studio."
```
simonvanderveldt: We should probably change this to something like "Multi-platform music-creation system for production, performance and DJing" (from their website)

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
mrVanDalo: DONE

mrVanDalo:
Yeah, I'll add back libav as option and ffmpeg as option as well.

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```diff
@@ -0,0 +1,102 @@
...
+	x11-libs/libXrender
+	x11-libs/libdrm
+	x11-libs/libxcb
+	x11-libs/libxkbfile
+	x11-libs/pixman
+	x11-libs/xcb-util-wm
+	virtual/opengl
+	virtual/udev
+	jack? ( virtual/jack )
+	libav? ( media-video/libav )
```
simonvanderveldt: Close, but no cigar ;)
You only need to depend on `virtual/ffmpeg`, that's it, no USE flag or conditionals necessary.
See the comments in the ffmpeg virtual ebuild https://github.com/gentoo/gentoo/blob/master/virtual/ffmpeg/ffmpeg-9-r2.ebuild

simonvanderveldt:
> bitwig2 would get a new slot. I only have a license for bitwig1.

Just thought about this morning. Was wondering how we'd do this?
Install to `/opt/bitwig-studio1` and name the symlink `/usr/bin/bitwig-studio1`?

media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
mrVanDalo: I see

mrVanDalo:
Yeah, that would be a good choice. It might be a good Idea to install the new bitwig to bitwig-studio2.
Hmm did not thought of that for now, but good point.

mrVanDalo:
true. I think it is better not to use slots, I don't want to type `bitwig-studio1` or `bitwig-studio2` to start them.

mrVanDalo:
I fixed the ffmpeg stuff and put it in slot=0. so bitwig1 and bitwig2 can't be installed at the same time.

PR #5, opened 2017-05-23
Add CI

simonvanderveldt:
This is similar to the setup I'm already using for my personal overlay, it does a `repoman full` check for every PR.
Also it allows us to automatically check for new versions of git based projects using a very simply Python script I wrote https://github.com/simonvanderveldt/newversionchecker
If you want I can move or fork newversionchecker to the gentoo-audio org.

See https://travis-ci.org/simonvanderveldt/simonvanderveldt-overlay/builds/227322743?utm_source=github_status&utm_medium=notification for an example of the repoman check and https://github.com/simonvanderveldt/simonvanderveldt-overlay/pull/9 and this PR for an example of the feedback to a PR.

See https://travis-ci.org/simonvanderveldt/simonvanderveldt-overlay/builds/235029235 for an example of the output and https://github.com/simonvanderveldt/simonvanderveldt-overlay/issues/12 for an automatically created version bump issue.

For now I've disabled the version checks because we don't have any packages yet :P

Eventually I'd like to add actual merging of the package that was added/changed in a PR with the default USE flags, but that needs a bit of work.
Also the version check should get the repo's URL from the ebuilds in the repo instead of using a separate list in the config file, but I haven't had the time to determine how I can properly pull that info from portage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Audio overlay
+# Audio overlay [![Build Status](https://travis-ci.org/gentoo-audio/audio-overlay.svg?branch=master)](https://travis-ci.org/gentoo-audio/audio-overlay)
 
 Overlay containing pro audio applications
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -24,3 +24,9 @@ If you run into problems please [create an issue](https://github.com/gentoo-audi
 ## Automated quality control
 - GitHub's [branch protection](https://help.github.com/articles/about-protected-branches/) is enabled for the `master` branch
 - Changes can only be done using [pull requests](https://help.github.com/articles/about-pull-requests/) and need at least one approval
+- Pull requests can only be merged if they pass the automated tests, which are run by [Travis CI](https://travis-ci.org)
```
mrVanDalo: would be nice to point the links here to our project, not to the sales page.

mrVanDalo:
I like it.

README.md
simonvanderveldt: Which link would you want to go where exactly?

PR #23, opened 2017-06-10
Add media-sound/drumkv1

### LINKED ISSUE #20: Add drumkv1

simonvanderveldt:
Initial stab at adding drumkv1, both the 9999/live from `master` ebuild as well as 0.8.2.
One issue I ran into is when compiling whilst an older version was installed I got multiple undefined references, like http://www.rncbc.org/drupal/comment/7452#comment-7452
Had to uninstall the old version to make it work. Not sure if this is really a problem or not though.

This uses the `9999` ebuild as the basis and the current release 0.8.2 is a symlink to that ebuild.
Inside the ebuild are some `if` statements to differentiate between the live and release ebuilds.
This prevents code duplication. If stuff in `master` starts working differently the symlink can be replaced by a copy.
Whilst so far I like doing it this way @mrVanDalo already said he's not really a fan of it.
@marcan What's your opinion on this?

Two things I'm not 100% sure about:
- The USE flag name for the standalone application/executable. The proaudio overlay had a couple of packages that used `standalone`, I've adopted that. It properly describes what the USE flag does imho
- The default enabled USE flags. Since the standalone version is the preferred one I've gone for `+standalone`. Since at least one MIDI connection is required to really use the standalone application I've also gone for `+jackmidi` since JACK should be preferred over ALSA

And finally this also includes adding drumkv1 to the list of repos to check so we get an automatic version bump issue when a new release/tag is created.

Let me know what you guys think.

Fixes #20


.travis.yml
```diff
@@ -15,4 +15,4 @@ cache:
 
 script:
   - 'if [ "${TRAVIS_EVENT_TYPE}" != "cron" ]; then docker run --rm -ti -v "${HOME}"/.portage:/usr/portage -v "${PWD}":/overlay -w /overlay gentoo/stage3-amd64:latest /overlay/tests/runtests.sh; fi'
-  # - 'if [ "${TRAVIS_EVENT_TYPE}" == "cron" ]; then docker run --rm -ti -v "${PWD}/tests/newversionchecker.toml":/app/newversionchecker.toml -e GITHUB_API_TOKEN simonvanderveldt/newversionchecker; fi'
+  - 'if [ "${TRAVIS_EVENT_TYPE}" == "cron" ]; then docker run --rm -ti -v "${PWD}/tests/newversionchecker.toml":/app/newversionchecker.toml -e GITHUB_API_TOKEN simonvanderveldt/newversionchecker; fi'
```
mrVanDalo: this belongs in a separated commit, It has nothing to do with media-sound/drumkv1

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,82 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
```
mrVanDalo: deprecated, plz remove

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	if [[ ${PV} == *9999 ]]; then
+		eautoreconf
+	fi
+
+	# Disable stripping
+	echo "QMAKE_STRIP=" >> src/src_core.pri.in
+	echo "QMAKE_STRIP=" >> src/src_jack.pri.in
+
+	local -a myeconfargs=(
+		$(use_enable debug)
```
mrVanDalo: I find the `use_enable` very hard to read, this is why there is `usex`
```
--enable-debug=$(usex debug)
--enable-qt4=$(usex qt4)
```
is much more understandable. Of course when you only use `use_enable` this does not look that obvious, but if there is `use_with` `use_without` , ... it gets hard to read very quickly. 

Here for example, it's not that obvious what is going on

```
$(use_enable alsa alsa-midi)
```

but here it is :

```
--enable-alsa-midi=$(usex alsa)
```

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,82 @@
...
+"
+DEPEND="${RDEPEND}"
+
+src_configure() {
+	if [[ ${PV} == *9999 ]]; then
+		eautoreconf
+	fi
+
+	# Disable stripping
+	echo "QMAKE_STRIP=" >> src/src_core.pri.in
```
mrVanDalo: why disable stripping ?

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	echo "QMAKE_STRIP=" >> src/src_jack.pri.in
+
+	local -a myeconfargs=(
+		$(use_enable debug)
+		$(use_enable qt4)
+		$(use_enable standalone jack)
```
mrVanDalo: The flags are confusing. 
`standalone` doesn't tell anything whats going on.
Also the dependencies show that the real flag here is `jack`

marcan:
I can't repro the linking issue, it seems to me g++ looks in . before /usr/lib for the relevant library. If you have an example log of the build failure having an older version installed, can you post it? Trying 0.8.2 here. If it's still an issue, it should be fixed.

I've never seen the symlink approach per se, but having conditional logic to share code between the -9999 and release ebuilds is common. I'm not opposed to the symlink approach (unless there's a downside I don't know about). It's easy to make a copy later if we need to diverge.

I think `USE=standalone` is fine. It's not about JACK support per se, it's about standalone builds (as opposed to only an LV2 plugin) that happen to use JACK as a backend. USE flags aren't about dependencies, they're about features; `standalone` describes the feature. However, this needs to go into `profiles/use.local.desc` with a description (or perhaps `use.local` if we expect enough packages in this overlay to use it; it feels like it might be a common enough pattern). Same goes for every other USE flag that isn't in the upstream use.desc.

Is there actually a reason for the `jackmidi` USE flag? If the JACK dep is required for standalone builds anyway (and MIDI support is irrelevant in lv2 mode), why not always enable jackmidi in them? Same with `jacksession`. Unless there's a use case for *disabling* support for these things, might as well always enable them if they add no further deps.

The typical meaning of the `qt4` and `qt5` USE flags is that each one enables the respective mode. E.g. on my system, both are set globally. I think the package should have both flags and `REQUIRED_USE="^^ ( qt4 qt5 )"` to force the user to pick one in `package.use` if necessary.

mrVanDalo:
> I think USE=standalone is fine. It's not about JACK support per se, it's about standalone builds (as opposed to only an LV2 plugin) that happen to use JACK as a backend. 

Than this confuses me a bit : 
```
standalone? ( virtual/jack )
￼alsa? ( media-libs/alsa-lib )
```

If there is a `jackmidi` flag than the `alsa` flag should be `jackmidi`

But to reveal my real intentions I agree in the number of flags I would reduce the flags to the following

```
IUSE="-debug -qt4 +jack alsa lv2 osc"
or 
IUSE="-debug -qt4 +standalone alsa lv2 osc"
```
Btw I would use `debug` only for the `9999` builds

media-sound/drumkv1/drumkv1-9999.ebuild
simonvanderveldt: Because stripping shouldn't be done during install, see

From https://www.gnu.org/prep/standards/html_node/Standard-Targets.html
> Do not strip executables when installing them. This helps eventual debugging that may be needed later, and nowadays disk space is cheap and dynamic loaders typically ensure debug sections are not loaded during normal execution.

This allows everyone to build a proper debug build if necessary, see https://wiki.gentoo.org/wiki/Debugging#Don.27t_strip_symbols

media-sound/drumkv1/drumkv1-9999.ebuild
simonvanderveldt: I'll check if there's a policy for this with upstream.
AFAIK it's use the `use_*` helpers where possible, only if that doesn't work use `usex`
We could choose our own policy here, but to me it's just a lot more typing for something that portage already does for.

media-sound/drumkv1/drumkv1-9999.ebuild
mrVanDalo: Ok but than it's better to use the `debug` flag for it

```
if use debug ; then 
echo "QMAKE_STRIP=" >> src/src_core.pri.in
echo "QMAKE_STRIP=" >> src/src_jack.pri.in
fi
```

simonvanderveldt: No, stripping is something portage should do, not the package's install script

mrVanDalo: I see, but shouldn't you prevent portage from stripping the binaries when using the debug flag ?

simonvanderveldt:
> I can't repro the linking issue, it seems to me g++ looks in . before /usr/lib for the relevant library. If you have an example log of the build failure having an older version installed, can you post it? Trying 0.8.2 here. If it's still an issue, it should be fixed.

I think I should be able to reproduce this, will post here when I do

> I've never seen the symlink approach per se, but having conditional logic to share code between the -9999 and release ebuilds is common. I'm not opposed to the symlink approach (unless there's a downside I don't know about). It's easy to make a copy later if we need to diverge.

:+1: 

> I think USE=standalone is fine. It's not about JACK support per se, it's about standalone builds (as opposed to only an LV2 plugin) that happen to use JACK as a backend. USE flags aren't about dependencies, they're about features; standalone describes the feature. 

Exactly, it's about what feature/functionality is enabled.
Agree with @mrVanDalo though that the current combo of USE flags doesn't make sense. JACK MIDI and JACK session don't introduce any new dependencies so they don't necessarily need a USE flag.
The reason I added the JACK MIDI one was because drumkv1 supports both ALSA and JACK MIDI.
Do we just want to go for an `alsa` USE flag as optional and that's it?

> However, this needs to go into profiles/use.local.desc with a description (or perhaps use.local if we expect enough packages in this overlay to use it; it feels like it might be a common enough pattern). Same goes for every other USE flag that isn't in the upstream use.desc.

Are you sure about this? I remember trying to add some "custom" USE flags to the global settings of my own overlay and I believe either portage or repoman complained that it should be set in `metadata.xml`.

> The typical meaning of the qt4 and qt5 USE flags is that each one enables the respective mode. E.g. on my system, both are set globally. I think the package should have both flags and REQUIRED_USE="^^ ( qt4 qt5 )" to force the user to pick one in package.use if necessary.

I looked this up before making the ebuild and afaik where possible the use of `REQUIRED_USE` should be prevented.
This seems to be the relevant situation for drumkv1 https://wiki.gentoo.org/wiki/Project:Qt/Policies#Optional_support_for_one_of_two_Qt_versions
It seems like I didn't get it entirely right though.
I'll add a `qt5` USE flag and make sure `qt5` is preferred when both are set

media-sound/drumkv1/drumkv1-9999.ebuild
simonvanderveldt: > I see, but shouldn't you prevent portage from stripping the binaries when using the debug flag ?

No, that's not something the `debug` USE flag should determine

`FEATURE=nostrip` is a portage thing, which you could set globally, once when emerging or per package in portage's `env.d`

`USE=debug` is a package thing that enables/disables creating a debug build for that package (or globally if you'd set it in your `make.conf`)

simonvanderveldt:
I've pushed some changes in a separate commit. I don't think I've addressed all issues, but it should be better.
@marcan Does the qt4/qt5 stuff look OK to you like this?
@mrVanDalo the USE flags should now match your suggestion except for the addition of an explicit `qt5` USE flag.

simonvanderveldt:
One small question, do we want to prevent/use one of the L10N/LINGUAS options to determine if translations should be installed?

```
$ ls -ahl /var/tmp/portage/media-sound/drumkv1-0.8.2/image/usr/share/man/man1 
total 26K
drwxr-xr-x 2 root root   4 Jun 10 11:44 .
drwxr-xr-x 3 root root   3 Jun 10 11:44 ..
-rw-r--r-- 1 root root 596 Jun 10 11:44 drumkv1.1.bz2
-rw-r--r-- 1 root root 771 Jun 10 11:44 drumkv1.fr.1.bz2
```

media-sound/drumkv1/drumkv1-9999.ebuild
marcan: I'm fine with `use_enable`; yes it takes some getting used to but it's the standard, no reason to diverge.

marcan:
Doesn't the lv2 only build require Qt? I assume it still builds a GUI. Thanks for finding that Qt policy doc; to me it seems our case is 'Requires one of two Qt versions', which means we need to drop the qt4 flag outright from IUSE and REQUIRED_USE shouldn't mention any qt flags.

We should check the localization settings to control installation of translations, yes, though that isn't a huge priority IMO (impact is minor).

+1 on the other comments.

simonvanderveldt:
> Doesn't the lv2 only build require Qt? I assume it still builds a GUI. Thanks for finding that Qt policy doc; to me it seems our case is 'Requires one of two Qt versions', which means we need to drop the qt4 flag outright from IUSE and REQUIRED_USE shouldn't mention any qt flags.

You're right on both accounts, I'll push a new commit that addresses this

> We should check the localization settings to control installation of translations, yes, though that isn't a huge priority IMO (impact is minor).

I'll have a quick look how easy this is, have never done anything with both `LINGUAS` and `L10N` before. If it's easy I'll include it, otherwise I'm OK with it being an improvement for later.

mrVanDalo:
> I've pushed some changes in a separate commit. Will squash once everything is OK/approved.
I don't think I've addressed all issues, but it should be better.

Well this is a feature branch so a force push does no harm. Also you should split this pr in 2 commits 
1 for the drumkitv1 and one for the travis updates
https://wiki.gentoo.org/wiki/Gentoo_GitHub#Separate_.28or_.22atomic.22.29_commits

media-sound/drumkv1/drumkv1-9999.ebuild
mrVanDalo: All packages I pushed into upstream people told me to use `usex` instead of `use_enable` because it is better to read. And I must agree with them. The syntax of `use_enable` `use_with` must always be looked up, and these commands safe you just 3 minus symbols and equal symbol.

mrVanDalo: > it's the standard, no reason to diverge.

sorry simply not true

media-sound/drumkv1/metadata.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>simon.vanderveldt+gentoo-overlay@gmail.com</email>
+		<name>Simon van der Veldt</name>
+	</maintainer>
+  <use>
+		<flag name="standalone">Enable building of standalone application</flag>
```
marcan: There seems to be some tabs/spaces mixing going on in this file.

media-sound/drumkv1/drumkv1-9999.ebuild
marcan: Well:

```bash
$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep '\buse_enable\b' | wc -l    
19272
$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep -- '--enable-.*usex' | wc -l
51
```
If there's a written policy that says we should use `usex` I'm all for it, but failing that it seems `use_enable` and friends are overwhelmingly the standard.

mrVanDalo:
```
13016:07 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=1 | wc -l
1
16:08 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=2 | wc -l
344
16:08 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=3 | wc -l
165
16:08 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=4 | wc -l
1734
16:08 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=5 | wc -l
16776
16:08 palo@workhorse [portage]$ find /usr/portage -name '*.ebuild' 2>|/dev/null | xargs grep EAPI=6 | wc -l
10915
```
same holds for EAPI=6 vs rest

mrVanDalo: @marcan this is not a very important topic to me, at the end its up to the maintainer to decided, just wanted to point out `use_enable` is always very confusing. 

Also you query for `usex` does not seem correct, u only query for `--enable` you are missing `with` `without` `disable` and other build tools who use it.

marcan: My query was for `use_enable` which maps to `usex` with `--enable`. If you want to compare all variants then you'd have to expand both queries.

I fail to see how the EAPI level has anything to do with this. As something that increments over time, it's natural that the latest EAPI wouldn't necessarily be the most popular yet. There's a big difference between the orders of magnitudes involved anyway.

mrVanDalo:
```
use_enable : Echoes either --enable-str=val or --disable-str depending upon useq flag
```
so at least the query should take into account `--disable`

With the `EAPI` I wanted to pointed out that you cant derive a standard/best-practise from the usage.

marcan: Even if you count --enable-x and --disable-x with a very generous regex, you still only get 62 hits.

Do you have upstream examples where `usex` is increasingly preferred over `use_enable` and friends? As I said, I'm not opposed to it, but it feels wrong to go against the grain without some evidence. Mostly because I don't want to wind up reverting it if we're told to use use_enable after all if we try to upstream :)

mrVanDalo: Yeah just where you ask, I realized I mixed it with the `cmake-utils_use_enable`.

Here I can give you an example, in supercollider I had this : 
```
$( cmake-utils_use_disable !sndfile libsndfile )
```
which I turned (after a long time of work) to this : 
```
-DNO_LIBSNDFILE=$(usex !sndfile)
```
Everybody instantly understands the second one, but the first one is a riddle.

marcan: Yeah, I agree that in the cmake case it's more obscure and trying too hard to follow an existing convention for little benefit.

simonvanderveldt:
I think I've address all comments so far :)
Also enabled the "rebase merge" option for this repo, otherwise we'd still end up with a single squashed commit.

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,77 @@
...
+HOMEPAGE="http://drumkv1.sourceforge.net/"
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3 autotools
+	EGIT_REPO_URI="https://github.com/rncbc/${PN}.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://download.sourceforge.net/${PN}/${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+LICENSE="GPL-2"
```
marcan: Should be GPL-2+

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,77 @@
...
+SLOT="0"
+RESTRICT="mirror"
```
marcan: Why restrict mirroring if it's GPL?

media-sound/drumkv1/drumkv1-9999.ebuild
```diff
@@ -0,0 +1,77 @@
...
+inherit fdo-mime
+
+DESCRIPTION="An old-school all-digital drum-kit sampler synthesizer with stereo fx"
+HOMEPAGE="http://drumkv1.sourceforge.net/"
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3 autotools
+	EGIT_REPO_URI="https://github.com/rncbc/${PN}.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://download.sourceforge.net/${PN}/${P}.tar.gz"
```
marcan: Portage knows about SourceForge mirrors. The syntax is `mirror://sourceforge/${PN}/${P}.tar.gz`.

simonvanderveldt: Good point that you mention this. It was actually something I wanted to discuss.

I started with the releases on GitHub, but they are just GitHub's auto generated tarballs, which means no `./configure` script, so you have to run `eautoreconf` just like when cloning the project.
I then switched to sourceforge because the releases there did have a `./configure` script.

Now I was wondering what we'd prefer to do:
- Use the sourceforge mirror and only call `eautoreconf` for the `9999` ebuild
- Use GitHub's release and always call `eautoreconf`

simonvanderveldt:
I ran into the same issue as with Bitwig #11, i.e. no high res icon in GNOME's tab switcher. I'll see if I can fix it.

media-sound/drumkv1/drumkv1-9999.ebuild
marcan: I don't think it makes a huge difference, but I'm inclined to go with SourceForge because those seem to be the intended "final" tarballs. If we find in the future that e.g. upstream starts neglecting to release the SF tarballs in a timely fashion even though releases are tagged on GitHub, we could switch.

simonvanderveldt: That makes sense, I'll fix it right away

simonvanderveldt:
@marcan both issues regarding mirroring/src/license metadata should now be fixed

> I ran into the same issue as with Bitwig #11, i.e. no high res icon in GNOME's tab switcher. I'll see if I can fix it.

I'm not going to try to fix this, it's a bit too much to sed/patch. I've filed an issue rncbc/drumkv1#16, let's see what Rui says first :)

Also, regarding the french man page, I'll leave that as-is for now as well. Couldn't really figure out what the correct way of fixing that would be and like @marcan already said the impact is minor

marcan:
I think you missed the `LICENSE="GPL-2"` -> `GPL-2+` change :)
Everything else looks good! Feel free to fix that and hit the merge button.

simonvanderveldt:
Thanks for the feedback! The ebuild has become a lot better because of it.
I'll use it a template for synthv1 next.

PR #25, opened 2017-06-11
Tests: repoman dont run xmlparse, include dev profiles

simonvanderveldt:
Small change, drops xmlparse because it doesn't actually do anything. See https://github.com/gentoo/portage/blob/a514908c7fd29e668dd775d7a32c949dbc8fe7d7/repoman/pym/repoman/argparser.py#L118. The `xml_parse` variable isn't used anywhere in repoman's codebase
Adding `-d` doesn't do anything in our case but gets rid of the unnecessary `Note: use --include-dev (-d) to check dependencies for 'dev' profiles` output

### DIFF
--- a/tests/runtests.sh
+++ b/tests/runtests.sh
@@ -9,4 +9,4 @@ emerge --sync
 emerge -q --buildpkg --usepkg dev-vcs/git app-portage/repoman
 
 # Run the tests
-repoman full --xmlparse
+repoman full -d


tests/runtests.sh
```diff
@@ -9,4 +9,4 @@ emerge --sync
 emerge -q --buildpkg --usepkg dev-vcs/git app-portage/repoman
 
 # Run the tests
-repoman full --xmlparse
+repoman full -d
```
mrVanDalo: From `man repoman`
```       -d, --include-dev
              Include dev profiles in dependency checks.
```
I never fully understood what a dev profile is. I thought this will call repoman on the dependencies as well.

simonvanderveldt: AFAIK it's for running tests using the `default/linux/amd64/*/developer` profiles.
Could be wrong though.

PR #26, opened 2017-06-11
Tests: emerge new or changed ebuild(s)

simonvanderveldt:
This updates the tests run for PRs to also emerge the new or changed ebuilds from that PR.


tests/runtests.sh
```diff
@@ -10,3 +12,10 @@ emerge -q --buildpkg --usepkg dev-vcs/git app-portage/repoman
 
 # Run the tests
 repoman full --xmlparse
+
+# Merge the new or changed ebuild(s)
+EBUILDS=($(git diff --name-only --diff-filter=d "${BASE_BRANCH}" | grep ".ebuild"))
+for EBUILD in "${EBUILDS[@]}"
+do
+  repoman merge "${EBUILD}"
```
mrVanDalo: what is `merge` doing ? can't find it in `man repoman`

simonvanderveldt:
@marcan @mrVanDalo @domichel What do you think of this? https://travis-ci.org/gentoo-audio/audio-overlay/builds/246744404?utm_source=github_status&utm_medium=notification

I've added the setbfree commits so it would run some actual ebuild/emerge tests, will drop that commit if/once this PR is OK.
Travis now runs two tests in a so called matrix (which also means in parallel).
- The repoman check we already had
- `emerge` of all ebuilds added or changed vs the branch the PR is going to be merged to (`master` most of the time)

tests/emerge.sh
```diff
@@ -0,0 +1,35 @@
...
+# Disable news messages from portage and disable rsync's output
+export FEATURES="-news" PORTAGE_RSYNC_EXTRA_OPTS="-q"
+
+# Update the portage tree and install dependencies
+emerge --sync
+emerge -q --buildpkg --usepkg dev-vcs/git
+
+# Setup overlay
+mkdir -p /etc/portage/repos.conf
+echo "[audio-overlay]" > /etc/portage/repos.conf/localrepo.conf
```
mrVanDalo: just an alternative 
```bash
cat > /etc/portage/repos.conf/localrepo.conf <<EOF
[audio-overlay]
location = /usr/local/portage
EOF
```

simonvanderveldt: Yeah, I was lazy here :P
Will update it

simonvanderveldt:
> Maybe it is simpler to use ebuild instead of a emerge for the builds. Less setup, but maybe we miss some repository changes.

I started with `ebuild` but that doesn't install dependencies, so we have to use `emerge` here.
I already split the emerge in 2 parts, 1 to install the dependencies and 2 to install the package itself. That second part could use `ebuild`, but that doesn't really gain us anything.

I tried to keep the setup as simple as possible. I can image in the future we might need package specific options for dependencies, which means just env vars won't be enough.

One thing I'm considering is running the script that determines which ebuild are new/changed first and the pass the list of ebuilds into the container. That way it wouldn't be necessary to install git explicitly during the emerge test setup. Still it would need to be installed for almost every 9999 ebuild anyway, so not sure if it really makes sense. I don't really like having to pass info around, tends to be a fragile thing to do.

simonvanderveldt:
After some consideration I did split the part that determines which ebuilds are new or changed from the emerge test, so that the test will really show all dependencies that are being installed.

I think we can now merge this and start making use of it :)

Issue #24, opened 2017-06-11
Add proaudio specific profile?


simonvanderveldt:
Follow up from #9. Shall we add a pro audio specific profile? And if so, what should it do?

mrVanDalo:
> Honestly I don't see why proaudio profiles should have anything to do with init system or desktop environment choice. Just enable audio-related flags and that's it. Going into init systems and desktops is making a lot of unnecessary assumptions. Desktop-specific variants should just inherit from the standard desktop profiles.

I'm on the side of @marcan.

domichel:
> IMHO if we want an audio-specific profile, we need to start from scratch and consider what should be in there, starting with the basics like jack and lv2.
You are right, that's fine for me.

> Desktop-specific variants should just inherit from the standard desktop profiles
We must keep in mind than in portage, we have
`/usr/portage/profiles/targets/desktop/gnome/` and `/usr/portage/profiles/targets/desktop/plasma/`, both with parent pointing to `/usr/portage/profiles/targets/desktop/`.

That imply we must have 3 profiles: `profiles/targets/desktop/gnome/`, `/usr/portage/profiles/targets/desktop/plasma/` and `profiles/targets/desktop/`, each of them with the audio flags and pointing to the related portage profile. If symlinks do works here, we can even maintain the USE flags only into the desktop profile, and use a symlink or 2 into the plasma and gnome profiles,,

domichel:
I just made a pull request, so you can review it.

simonvanderveldt:
I did some thinking on this and I think the initial question "Shall we add a pro audio specific profile?" is something we jumped over too fast.

We should determine what the benefits to the users would be and what would need to be done to make it possible and guarantee to the users that it keeps working.

From the discussion in #28 I think we can conclude the only thing it would bring would be some predefined USE flags.
Looking at #28 a serious amount of files is needed to make it work. Which also means we'd need to somehow guarantee that all of them would work.

My suggestion would be to not add profiles to the overlay. I don't think it's worth the time/effort just to add a couple of USE flags. The USE flags should be a user's own choice anyway.

What we can (and probably should anyway) do is document properly, ideally on the gentoo wiki, what a pro audio/audio production setup should consist of and have a suggestion for useful USE flags there.

mrVanDalo:
I use Funtoo, so it will not affect me.

marcan:
[...]

PR #28, opened 2017-06-13
Profiles

domichel:
Audio pro set of profiles. It include profiles for desktop, gnome and plasma with parents pointing to the related portage profiles. That for 13.0 and both x86 and amd64.

The profiles/default hierarchy use symlinks to profiles/targets, which imply all the maintenance will be done in profiles/targets/desktop/proaudio/

To make the same for dev would be pretty simple (copy 13.0/* as dev/* into the default hierarchy and make the parent files to point to the corresponding dev profiles).


mrVanDalo:
do these profiles interfere with my local set profiles ? I mean when I selected `desktop` in `epro` as flavor, will these profile settings be mixed, when load from this repository ?

simonvanderveldt:
Do we really need 44 files to add 3 profiles? (I never really got into how profiles are created, but this seems like a lot)

mrVanDalo:
It seems my funtoo uses a bit different profile concept than gentoo.
http://www.funtoo.org/Funtoo_Profiles

domichel:
> Do we really need 44 files to add 3 profiles? (I never really got into how profiles are created, but this seems like a lot)

I know. The issue to solve is than in portage, we have things like desktop/gnome/systemd with parent to desktop/gnome with parent to desktop. For a profile to work in an overlay with portage, we must use the same hierarchy. That because a desktop profile into an overlay can have only one parent. The alternative is to develop and maintain a complete desktop profile. But it is much simpler to use the existing portage profiles as parent for the overlay profiles.

Also, if we have only one desktop profile, users using a gnome or plasma profile will not be able to use it, as its parent will be desktop in portage.

As we don't want to deal with other stuffs than audio, we will add the same flags audio in all the profiles. That imply we can use symlinks into the profile overlay, with the advantage to have only a few files to maintain into the targets hierarchy.

The profiles of the proaudio overlay works that way. I remember it was very time consuming to get them to work well with portage. I was trying a lot of ways, but only that was working fine.

profiles/ChangeLog
```diff
@@ -0,0 +1,21 @@
...
+  - Updated eapi to 5
+  - Dropped the outdated and in portage desktop profile USE flags
+
+  13 Jun 2017; Dominique Michel <dominique_libre@users.sourceforge.net>
+  - Removed non mandatory USE flags
+  - Added desktop/gnome and desktop/plasma with parents to the corresponding
+    portage profiles
+
+  14 Jun 2017; Dominique Michel <dominique_libre@users.sourceforge.net>
+  - Add profiles.desc: needed by eselect
```
marcan: ChangeLogs are dead. This is a Git repository.

profiles/targets/desktop/proaudio/make.defaults
```diff
@@ -0,0 +1,6 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
```
marcan: No header lines, those are obsolete.

profiles/use.desc
```diff
@@ -0,0 +1 @@
+lv2 - Add support for Ladspa V2
```
marcan: Nobody calls it Ladspa V2 (even though that's where the name comes from), and this doesn't say much about what the flag actually does. Should be something like "Add support for LV2 plug-ins"

profiles/default/linux/amd64/13.0/desktop/proaudio/gnome/parent
```diff
@@ -0,0 +1 @@
+/usr/portage/profiles/default/linux/amd64/13.0/desktop/gnome
```
marcan: I don't like this, with a hardcoded path. I figured there must be a better way of doing this, and indeed there is. `man portage` says:

> If layout.conf is new enough, you can also use the \<repo\>:\<path\> syntax. The \<repo\> is the same string as is stored in the repo_name file (or omitted to refer to the current repo), and \<path\> is a subdir starting at  profiles/.

That makes a lot more sense.

marcan:
Higher level comment: we need to consider whether we really need or want to mirror the full desktop profile hierarchy in proaudio. Note that using a desktop profile (or a specific sub-profile) is not at all required to use said desktop. For example, I'm a perfectly happy Plasma user with:

    $ eselect profile show
    Current /etc/portage/make.profile symlink:
      hardened/linux/amd64

And I'm unlikely to actually use the proaudio profile myself for that reason (unless we add a hardened/linux/amd64/proaudio profile, but that sounds silly). Just like the hardened team doesn't bother maintaining desktop sub-profiles, we need to consider whether we should do really do the same with proaudio.

Ultimately it just comes down to user convenience. None of this is required to actually use audio packages, or to use a given desktop. We're just talking about configuring USE defaults. So whether we want to mirror the entire desktop hierarchy or not is a decision we need to take.

Why do you say we can't use multiple parents? `man portage` doesn't say anything about that. That should cut down on the file count a lot, so we can have a single inheritable audio partial profile and then only have the `parent` and `eapi` files in each subdirectory. That avoids all the `package.use` and `make.defaults` symlinks.

profiles/ChangeLog
domichel: Removed

profiles/targets/desktop/proaudio/make.defaults
domichel: Fixed

profiles/use.desc
domichel: Fixed

profiles/default/linux/amd64/13.0/desktop/proaudio/gnome/parent
domichel: Unfortunately, it just doesn't work:
```
# repoman
!!! Unable to parse profile: '/etc/make.profile'
!!! ParseError: Parent '/var/lib/layman/proaudio/profiles/default/linux/amd64/13.0/desktop/proaudio/gentoo:default/linux/amd64/13.0/desktop' not found: '/var/lib/layman/proaudio/profiles/default/linux/amd64/13.0/desktop/proaudio/parent'
```
I get the same error with portage. On bugzilla, it is at least 3 repoman bugs related to PORTDIR and PORTDIR_OVERLAY. The consequence is we must use the old syntax.

domichel:
> Why do you say we can't use multiple parents? 

Because one profile of the overlay can have only one parent in portage.

> That should cut down on the file count a lot, so we can have a single inheritable audio partial profile and then only have the parent and eapi files in each subdirectory. 

That would be great but portage doesn't allow to mix profiles. In such a case, the only way is to make your own profile as explained here: https://wiki.gentoo.org/wiki/Profile_(Portage)#Combining_profiles

> So whether we want to mirror the entire desktop hierarchy or not is a decision we need to take.

If we make just one audio profile, the question is which portage profile will be its parent?
For user convenience to have the choice between the different desktop profiles is good. I think many, if not most, audio users will use a desktop profile. And with the symlinks, it is just a few files to maintain.
An alternative can be to have no profile at all.
A last one would be to mirror the entire profile hierarchy.

simonvanderveldt:
Closing as per the discussion in #24

PR #29, opened 2017-06-14
media-libs/libffado/libffado-2.3.0 and -9999: new package

marcan:
This is based on a cleaned up ebuild from proaudio.

There's a patch for 2.3.0 which is not upstream, but I haven't added it to -9999 because I'll be sending it upstream soon. Also, -9999 doesn't build right now, but that's due to (unlucky) upstream breakage, not the ebuild. Other than the PATCHES line, the ebuilds are identical.

Note that, unlike the proaudio ebuild, this ebuild does *not* depend on JACK (as that would introduce a circular dependency). Instead it blocks old versions of JACK. The JACK "dependency" is only about picking up the right API for JACK to use; I just always force it to the more recent one and block old releases.


marcan:
Looks like the problem fixed by the patch is already fixed upstream (in a different way), so no need to upstream that patch.

simonvanderveldt:
@marcan The purpose of adding this is to support ieee1394 in Jack, right? Or do you have other uses for it?

Do you think we can directly upstream this instead of adding it to the overlay first? And then combine it with an update of the jack-audio-connection-kit and jack2 ebuilds to add ieee1394 support?

marcan:
Correct. We could directly upstream it, but I figured we might as well review it and test it here before doing so. I intend it to be short-lived in the overlay.

My plan is to copy the latest JACK1/JACK2 ebuilds from portage to here, test them, then upstream. For the copy they would use the same ${PV} (since revbumping could conflict with a portage revbump); they would them be upstreamed as a proper revbump.

simonvanderveldt:
@marcan sounds good to me!

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+IUSE="debug qt4 +test-programs"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="dev-cpp/libxmlpp:2.6[${MULTILIB_USEDEP}]
+	dev-libs/dbus-c++
+	dev-libs/libconfig[cxx,${MULTILIB_USEDEP}]
+	media-libs/alsa-lib
+	media-libs/libiec61883[${MULTILIB_USEDEP}]
+	!<media-sound/jack-audio-connection-kit-0.122.0:0
+	!<media-sound/jack-audio-connection-kit-1.9.9:2
```
mrVanDalo: what about `media-sound/jack2` ? 

There is no  `media-sound/jack-audio-connection-kit-1.9.9:2` in gentoo upstream and `audio-overlay`

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	fi
+}
+
+src_prepare() {
+	default
+	multilib_copy_sources
+}
+
+multilib_src_configure() {
+	: # no-op
```
mrVanDalo: what is this ?

media-libs/libffado/libffado-2.3.0.ebuild
marcan: There was never an older version of media-sound/jack2. This is mostly for the benefit of those who might have had an older slotted jack2 from proaudio.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Says so right on the tin. It's a no-op to get rid of the default src_configure.

media-libs/libffado/libffado-2.3.0.ebuild
mrVanDalo: Ah I see.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Looks like this isn't actually necessary (maybe the scons eclass gets rid of it?) so I'll remove it (it came from the original ebuild).

media-libs/libffado/libffado-2.3.0.ebuild
simonvanderveldt: @marcan are these jack versions really incompatible? Or is it just something you didn't validate?
Asking because in the gentoo tree `jack-audio-connection-kit-0.121.3-r1` is the only one marked as stable.

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="6"
+
+PYTHON_COMPAT=( python2_7 )
+[[ "${PV}" = "9999" ]] && inherit subversion
```
simonvanderveldt: Can you move this into the `if [ "${PV}" = "9999" ]` block on line 14?
It doesn't really matter where `subversion` is inherited here and I'd prefer to keep the if/logic blocks to a minimum.

simonvanderveldt: Can you move this into the `if [ "${PV}" = "9999" ]` block on line 14?
It doesn't really matter where `subversion` is inherited here and I'd prefer to keep the if/logic blocks to a minimum.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: They aren't incompatible per se, but they require a different API which requires building libffado with different options (which is what it normally tries to autodetect but that's a horrible idea and the original ebuild had a circular dependency for it)... I could introduce a USE flag and alternate blocks depending on it, but it just feels like too much effort. I wasn't planning on adding libffado support to all JACK versions anyway, just the latest. Honestly, I expect people doing proaudio work to use ~arch for many packages.

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+HOMEPAGE="http://www.ffado.org"
+
+RESTRICT="mirror"
+if [ "${PV}" = "9999" ]; then
+	ESVN_REPO_URI="http://subversion.ffado.org/ffado/trunk/${PN}"
+	SRC_URI=""
+	KEYWORDS=""
+else
+	SRC_URI="http://www.ffado.org/files/${P}.tgz"
+	KEYWORDS="~amd64 ~ppc ~x86"
```
simonvanderveldt: Do we want to support more than `amd64`?
Do we even have people onboard who can guarantee it'll work/maintain it for other architectures?

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+[[ "${PV}" = "9999" ]] && inherit subversion
+inherit eutils python-single-r1 scons-utils toolchain-funcs udev multilib-minimal
+
+DESCRIPTION="Driver for IEEE1394 (Firewire) audio interfaces"
+HOMEPAGE="http://www.ffado.org"
+
+RESTRICT="mirror"
+if [ "${PV}" = "9999" ]; then
+	ESVN_REPO_URI="http://subversion.ffado.org/ffado/trunk/${PN}"
+	SRC_URI=""
```
simonvanderveldt: I think you can just drop this line

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	KEYWORDS=""
+else
+	SRC_URI="http://www.ffado.org/files/${P}.tgz"
+	KEYWORDS="~amd64 ~ppc ~x86"
+fi
+
+LICENSE="GPL-2 GPL-3"
+SLOT="0"
+IUSE="debug qt4 +test-programs"
```
simonvanderveldt: Is the `qt4` USE flag here for optional Qt4 support? Qt5 isn't supported?

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	else
+		myesconsargs+=(
+			BUILD_MIXER=false
+			BUILD_TESTS=false
+		)
+	fi
+	escons "${myesconsargs[@]}" "${@}"
+}
+
+src_unpack() {
```
simonvanderveldt: This one can be dropped, `src_unpack` from `subversion` already provides this automatically. See https://github.com/gentoo/gentoo/blob/master/eclass/subversion.eclass#L28

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,116 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="6"
+
+PYTHON_COMPAT=( python2_7 )
+[[ "${PV}" = "9999" ]] && inherit subversion
+inherit eutils python-single-r1 scons-utils toolchain-funcs udev multilib-minimal
```
simonvanderveldt: Do we have to support multilib for this package?

Or as a general question: Do we want to support multilib?
I have no experience with maintaining ebuilds for multilib, but it seems quiet a bit more involved than it is without it.

media-libs/libffado/libffado-2.3.0.ebuild
simonvanderveldt: @marcan OK, makes sense

simonvanderveldt:
@marcan Can you file a bug upstream/for gentoo about firewire support in JACK? That way then can warm up to the addition of a apcakge and we can start pushing these changes upstream asap.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Correct (as far as I know). This enables the mixer GUI. There's a bunch of movement in upstream SVN; I need to check to see if it's moving in the direction of supporting Qt5, but I do believe at least the stable version is still Qt4.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: I actually *do* have a PPC machine with firewire running Linux, though that's coincidental and I had no intention of actually running JACK on it (nor does it run Gentoo) :)

In general, do we want to support x86? I don't care personally, but other people might. For an overlay it seems reasonable to throw in ~x86; worst case something breaks and someone complains, best case we do nothing and it just works. Stuff breaking on x86 and not amd64 is rare.

marcan:
No symlink because 2.3.0 has a patch (without which the mixer doesn't work on my interface) :)

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Done.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Done.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Done.

media-libs/libffado/libffado-2.3.0.ebuild
simonvanderveldt: > I actually do have a PPC machine with firewire running Linux, though that's coincidental and I had no intention of actually running JACK on it (nor does it run Gentoo) :)

:P

> In general, do we want to support x86? I don't care personally, but other people might. For an overlay it seems reasonable to throw in ~x86; worst case something breaks and someone complains, best case we do nothing and it just works. Stuff breaking on x86 and not amd64 is rare.

Not sure, I prefer to be in control and automatically check if everything works so we don't have to rely on users to get feedback. So I'd be in favor of just supporting `amd64`.
If someone at some point asks if `x86` can be supported we can also just add it at that point and make him/her the maintainer :)

simonvanderveldt:
> No symlink because 2.3.0 has a patch

You could move the `PATCHES` variable declaration inside the `if [ "${PV}" = "9999" ]` block?

media-libs/libffado/libffado-2.3.0.ebuild
marcan: I think the answer is "probably nobody cares but getting rid of it is hard". The JACK client library definitely needs to be built with multilib support. That doesn't depend on ffado. But the JACK server is also a library and you can run a JACK server in-process. Does anyone actually do that? Do random things link to that library that might pull it in for 32-bit apps? Maybe we could just drop the drivers since surely nobody actually runs a 32-bit in process server on a 64-bit OS?

I don't know the answers to these questions, but it would involve messing with the JACK ebuild quite a bit more, and... that's probably more effort than just supporting multilib here.

marcan:
I don't think that's a good idea. The moment we have patches or per-version hacks involved we shouldn't be using symlinks; adding conditionals for individual versions (not a single binary -9999 vs rest) muddies everything. The -9999 ebuild should always be the "reference" version intended for a clean stable build or a 9999 live build, and then any customizations should be done on copies, IMHO.

media-libs/libffado/libffado-2.3.0.ebuild
marcan: Fair enough. Out it goes, let thy who still run x86 boxes complain, if any :)

media-libs/libffado/libffado-2.3.0.ebuild
simonvanderveldt: @marcan I'm fine with Jack is multilib so this has to be multilib as well ;)

simonvanderveldt:
> The moment we have patches or per-version hacks involved we shouldn't be using symlinks; adding conditionals for individual versions (not a single binary -9999 vs rest) muddies everything. The -9999 ebuild should always be the "reference" version intended for a clean stable build or a 9999 live build, and then any customizations should be done on copies, IMHO.

This sounds good to me as the basic policy.
I suggested it here because we just have a 9999 and a single versioned one, but if we'd add more versions we'd need to split them off anyway. Agreed that there shouldn't be other logic in there van 9999 vs rest :+1:

media-libs/libffado/libffado-2.3.0.ebuild
simonvanderveldt: OK, no problem, we can just leave it at Qt4 for now.

media-libs/libffado/libffado-2.3.0.ebuild
```diff
@@ -0,0 +1,104 @@
...
+	ESVN_REPO_URI="http://subversion.ffado.org/ffado/trunk/${PN}"
+	KEYWORDS=""
+else
+	SRC_URI="http://www.ffado.org/files/${P}.tgz"
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="GPL-2 GPL-3"
+SLOT="0"
+IUSE="debug qt4 +test-programs"
```
simonvanderveldt: Why is `test-programs` enabled by default? And shouldn't it be `test` as a USE flag?

media-libs/libffado/libffado-2.3.0.ebuild
mrVanDalo: test-programs should be enabled by default ?

audiodef:
I tried this ebuild, but for both jack2 and jack-audio-connection-kit, I get "FFADO driver was explicitly requested but could not be built."

simonvanderveldt:
@marcan I get the following QA notice when emergin this. Any idea how to fix it?

```
 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * src/rme/fireface_flash.cpp:613:64: warning: iteration 28u invokes undefined behavior [-Waggressive-loop-optimizations]
 * src/rme/fireface_flash.cpp:692:33: warning: iteration 28u invokes undefined behavior [-Waggressive-loop-optimizations]

 * Please do not file a Gentoo bug and instead report the above QA
 * issues directly to the upstream developers of this software.
 * Homepage: http://www.ffado.org
```

marcan:
@simonvanderveldt It fails on GCC6, but for a different reason. I'm adding a patch to fix that. I don't think we should be worrying about all the warnings for the ebuild; that's for upstream to fix. The QA warnings are worth noting upstream if they haven't fixed them yet, but I don't want to blindly touch that code unless we have a report of it actually causing a problem.

I tried this with JACK1 and it works fine (automagic dep or manually adding the flag), so I don't think there's anything left to do here other than the latest changes. OK to merge?

media-libs/libffado/libffado-2.3.0.ebuild
marcan: It's not really test suites but rather just standalone testing apps. I'll remove the default.

marcan:
I think we're mostly good here so I'll hit the merge button. Can always fix any further issues that arise later :)

PR #30, opened 2017-06-15
Add media-sounds/synthv1

### LINKED ISSUE #19: Add synthv1

simonvanderveldt:
Fixes #19


media-sound/synthv1/synthv1-9999.ebuild
```diff
@@ -0,0 +1,83 @@
...
+	media-libs/libsndfile
+	standalone? ( virtual/jack )
+	alsa? ( media-libs/alsa-lib )
+	lv2? ( media-libs/lv2 )
+	osc? ( media-libs/liblo )
+"
+DEPEND="${RDEPEND}"
+
+src_prepare() {
+	# Fix to ensure this desktop file is used to get the application's icon
```
simonvanderveldt: This works, for now it can stay here until it gets fixed upstream.
Then it would either be needed to move it into an if for !9999 or the ebuilds should be split

simonvanderveldt:
I've removed the symlink from 0.8.2 to 9999 because 0.8.2 is different from 9999.
The next release can be symlinked to the 9999 ebuild again.

media-sound/synthv1/synthv1-0.8.2.ebuild
```diff
@@ -0,0 +1,83 @@
+# Copyright 1999-2015 Gentoo Foundation
```
marcan: Header needs an update

media-sound/synthv1/synthv1-9999.ebuild
```diff
@@ -0,0 +1,76 @@
+# Copyright 1999-2015 Gentoo Foundation
```
marcan: Same here.

PR #32, opened 2017-06-18
media-sound/drumkv1: Fix low res icon, update copyright header

simonvanderveldt:
Small change, adds the same fix as I already did in #30 and as it has been fixed in upstream `master` to the 0.8.2 version of drumkv

### DIFF
--- a/media-sound/drumkv1/drumkv1-0.8.2.ebuild
+++ b/media-sound/drumkv1/drumkv1-0.8.2.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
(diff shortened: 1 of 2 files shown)


mrVanDalo:
It's hard to see what is new now. Did you just copy the file ?

simonvanderveldt:
@mrVanDalo Copied the file and added the `src_prepare`

media-sound/drumkv1/drumkv1-0.8.2.ebuild
```diff
@@ -1 +0,0 @@
-drumkv1-9999.ebuild
```
marcan: While you're here, mind fixing the copyright year? Seems we missed it last time.

simonvanderveldt: Sure, will do!

PR #36, opened 2017-06-23
Create media-sound/carla-9999.ebuild

audiodef:
KXStudio's audio plugin host.


marcan:
I made a bunch of comments to the Cadence ebuild and I think most of them apply here too.

marcan:
This is in the wrong directory too.

PR #37, opened 2017-06-23
Create media-sound/drumgizmo-0.9.12.ebuild

audiodef:
Drum track manipulation plugin for use with Ardour and other DAWs.

### DIFF
--- a/media-sound/drumgizmo-0.9.12.ebuild
+++ b/media-sound/drumgizmo-0.9.12.ebuild
@@ -1,0 +1,26 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+DESCRIPTION=""
+HOMEPAGE=""
+SRC_URI="http://www.drumgizmo.org/releases/${P}/${P}.tar.gz"
+
+LICENSE=""
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+DEPEND="media-libs/lv2 
+	x11-libs/libX11 
+	media-libs/libsndfile 
+	media-libs/zita-resampler 
+	dev-libs/expat 
+	media-libs/libsmf"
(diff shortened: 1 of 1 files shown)


media-sound/drumgizmo-0.9.12.ebuild
```diff
@@ -0,0 +1,26 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+DESCRIPTION=""
```
marcan: This needs a description and a homepage.

media-sound/drumgizmo-0.9.12.ebuild
```diff
@@ -0,0 +1,26 @@
...
+HOMEPAGE=""
+SRC_URI="http://www.drumgizmo.org/releases/${P}/${P}.tar.gz"
+
+LICENSE=""
```
marcan: Definitely needs a license.

media-sound/drumgizmo-0.9.12.ebuild
```diff
@@ -0,0 +1,26 @@
...
+	x11-libs/libX11 
+	media-libs/libsndfile 
+	media-libs/zita-resampler 
+	dev-libs/expat 
+	media-libs/libsmf"
+RDEPEND="${DEPEND}"
+
+src_configure()
+{
+	econf --enable-lv2
```
marcan: Should this be an `lv2` USE flag? Is it useful without lv2?

marcan:
This is in the wrong directory too.

media-sound/drumgizmo-0.9.12.ebuild
audiodef: Yes, but I wrote it this way because the cli works whether or not lv2 is enabled, and I wanted this to compile so it could be used with Ardour without bothering the user about it.

PR #38, opened 2017-06-23
media-sound/ebumeter-0.4.0.ebuild

audiodef:
Loudness measurement according to EBU-R128.

### DIFF
--- a/media-sound/ebumeter-0.4.0.ebuild
+++ b/media-sound/ebumeter-0.4.0.ebuild
@@ -1,0 +1,38 @@
+# Copyright 1999-2012 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=4
+inherit base toolchain-funcs
+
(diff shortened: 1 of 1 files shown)


media-sound/ebumeter-0.4.0.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2012 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
```
marcan: Header lines are legacy.

marcan:
This is in the wrong directory, and doesn't pass repoman once moved to the correct directory.

media-sound/ebumeter-0.4.0.ebuild
audiodef: I didn't write it. Feel free to change. I'm just passing along ebuilds I've been including in Gentoo Studio that aren't in Portage.

PR #40, opened 2017-06-23
media-sound/mididings: new package

marcan:


marcan:
Note that I dropped SMF support in the `99999999` ebuild because it now depends on pysmf, which is a separate package. I can add it if anyone cares about SMF support (but I don't).

marcan:
Also added slot operator deps to the library dependencies (`:=`). We should be doing this for all library dependencies, to ensure they get rebuilt when their SONAME change.

simonvanderveldt:
Rebased this PR on current master to kick off the emerge test.

@marcan did you deliberately name the ebuild `mididings-99999999`? I have never seen this, is there a reason not to use the regular four nines?

media-sound/mididings/mididings-20120419.ebuild
```diff
@@ -0,0 +1,54 @@
...
+REQUIRED_USE="|| ( alsa jack )"
+
+RDEPEND="${PYTHON_DEPS}
+	>=dev-libs/boost-1.34.1:=[python,threads,${PYTHON_USEDEP}]
+	dev-libs/glib:2=
+	dev-python/decorator[${PYTHON_USEDEP}]
+	alsa? ( media-libs/alsa-lib:= )
+	dbus? ( dev-python/dbus-python[${PYTHON_USEDEP}] )
+	inotify? ( >=dev-python/pyinotify-0.8[${PYTHON_USEDEP}] )
+	jack? ( >=media-sound/jack-audio-connection-kit-0.116.0 )
```
simonvanderveldt: Is mididings explicitly incompatible with jack2? If not this should be `virtual/jack`

media-sound/mididings/mididings-99999999.ebuild
```diff
@@ -0,0 +1,52 @@
...
+REQUIRED_USE="|| ( alsa jack )"
+
+RDEPEND="${PYTHON_DEPS}
+	>=dev-libs/boost-1.34.1:=[python,threads,${PYTHON_USEDEP}]
+	dev-libs/glib:2=
+	dev-python/decorator[${PYTHON_USEDEP}]
+	alsa? ( media-libs/alsa-lib:= )
+	dbus? ( dev-python/dbus-python[${PYTHON_USEDEP}] )
+	inotify? ( >=dev-python/pyinotify-0.8[${PYTHON_USEDEP}] )
+	jack? ( >=media-sound/jack-audio-connection-kit-0.116.0 )
```
simonvanderveldt: Is mididings explicitly incompatible with jack2? If not this should be `virtual/jack`
(same as above)

marcan:
@simonvanderveldt It's actually quite common, grep around for 99999999 in /usr/portage. It's necessary because live ebuilds have to be numerically higher than release versions, and this package (and similar ones) use a YYYYMMDD versioning scheme.

marcan:
I suspect the patches may have to do with how distros package boost. I'm not entirely sure what the behavior is in other distros; given the fact that the code for dealing with the suffixes exists I assume it works in at least some other environment. I don't know enough about how boost-python stuff is intended to know exactly who is at fault here.

simonvanderveldt:
> @simonvanderveldt It's actually quite common, grep around for 99999999 in /usr/portage. It's necessary because live ebuilds have to be numerically higher than release versions, and this package (and similar ones) use a YYYYMMDD versioning scheme.

Interesting, learned a new thing today :)
Seems like this is indeed how it's done for packages with datestamped ebuilds. Thanks for the pointer.

Looking at the emerge output on Travis, is it correct that mididings gets installed for multiple Python versions (2.7 and 3.4 in this case) and not just one?
(seems like this should actually be a pypi package, but apparently [it isn't](https://github.com/dsacre/mididings/issues/8))

marcan:
Yes, this is intentional and expected. Mididings is, after all, just a Python module. It gets built for whatever Python versions the user has configured. The actual script importing it could be targeted at any Python version installed.

marcan:
I always use "rebase" which does keep them separate :)

PR #48, opened 2017-06-23
media-sound/setbfree: new package

marcan:
Package-Manager: Portage-2.3.6, Repoman-2.3.2


media-sound/setbfree/setbfree-9999.ebuild
```diff
@@ -0,0 +1,53 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="6"
+inherit toolchain-funcs multilib
+
+DESCRIPTION="MIDI controlled DSP tonewheel organ"
+HOMEPAGE="http://setbfree.org"
+
+if [ "${PV}" = "9999" ]; then
+	inherit subversion git-r3
```
simonvanderveldt: @marcan I don't think subversion is correct here :)

media-sound/setbfree/setbfree-9999.ebuild
```diff
@@ -0,0 +1,53 @@
...
+else
+	SRC_URI="https://github.com/pantherb/setBfree/releases/download/v${PV}/${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="GPL-2+"
+SLOT="0"
+IUSE="convolution"
+
+RDEPEND="media-sound/jack-audio-connection-kit
```
simonvanderveldt: Unless setbfree only works with jack1 this should be `virtual/jack`

media-sound/setbfree/setbfree-9999.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	>=media-libs/alsa-lib-1.0.0:=
+	media-libs/liblo:=
+	media-libs/lv2
```
simonvanderveldt: Is it not possible to choose what to build (standalone application or lv2 plugin)/

media-sound/setbfree/setbfree-9999.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	convolution? ( media-libs/libsndfile:=
+		>=media-libs/zita-convolver-3.1.0:= )
+	media-fonts/dejavu"
```
simonvanderveldt: I believe this shoud be `virtual/ttf-fonts-1-r1`?
Or is it hardcoded to dejavu?

marcan: It actually used to be hardcoded to ttf-bitstream-vera, which is on its way out, so I changed it to dejavu (see later in the ebuild for the TTF path). It just uses a hardcoded TTF path, doesn't dynamically find a font :(

media-sound/setbfree/setbfree-9999.ebuild
marcan: Err. Yeah. Fixed.

media-sound/setbfree/setbfree-9999.ebuild
marcan: Fixed.

media-sound/setbfree/setbfree-9999.ebuild
marcan: It's a mess of automagic junk in a Makefile with no easy way to configure what goes in and what doesn't. As far as I can tell, the UI (even in standalone mode) depends on LV2. You can build the synth core without LV2 and the UI, but there's no way to force that (well, unless you do something ugly like set FONTFILE to a non-existent file to force the UI test to fail). And the synth core by itself is of limited use unless you already have all the presets you need (headless mode?) or are happy using MIDI to control everything.

stefantalpalaru:
The newest version is https://github.com/pantherb/setBfree/archive/v0.8.4.tar.gz .

marcan:
Welp, looks like upstream isn't using GitHub releases consistently (0.8.0 was the last available *release*).

Bumped to 0.8.4.

media-sound/setbfree/setbfree-9999.ebuild
simonvanderveldt: Seems like it still defaults to ttf-bitstream-vera https://github.com/pantherb/setBfree/blob/9c04792f1c16725eb978527add672eae2f938f26/common.mak#L6 :(

But I understand, not much we can then :(

PR #50, opened 2017-06-29
Distrho ports

audiodef:


media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+	media-libs/freetype
+	<dev-util/premake-4.3"
+
+RDEPEND=${DEPEND}
+
+src_prepare() {
+	"${WORKDIR}"/"${P}"/scripts/premake-update.sh linux
+}
+
+src_compile() {
```
marcan: I believe this src_compile and the src_install are the default, you should be able to just remove them.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+IUSE=""
+
+DEPEND="virtual/jack
+	media-libs/freetype
+	<dev-util/premake-4.3"
+
+RDEPEND=${DEPEND}
+
+src_prepare() {
+	"${WORKDIR}"/"${P}"/scripts/premake-update.sh linux
```
marcan: Probably don't need the `"${WORKDIR}"/"${P}"/` since that should be the current directory

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+
+IUSE=""
+
+DEPEND="virtual/jack
+	media-libs/freetype
+	<dev-util/premake-4.3"
+
+RDEPEND=${DEPEND}
```
marcan: You should quote `"${DEPEND}"`

This looks wrong though, surely dev-util/premake is only a build-time dep, not a runtime dep? Instead you should define RDEPEND first and then make DEPEND be RDEPEND plus your build deps, typically.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+# requires proaudio overlay
+
+EAPI=5
+
+inherit git-r3
+
+DESCRIPTION="Linux ports of Distrho plugins."
+HOMEPAGE="http://distrho.sourceforge.net"
+
+SRC_URI=""
```
marcan: This shouldn't be necessary to define.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2013 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# requires proaudio overlay
+
+EAPI=5
```
marcan: We require EAPI=6 for this overlay.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+
+IUSE=""
+
+DEPEND="virtual/jack
+	media-libs/freetype
+	<dev-util/premake-4.3"
```
marcan: Perhaps this should be a slot dep, like premake:3?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2013 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# requires proaudio overlay
```
marcan: Does this really? Why?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2013 Gentoo Foundation
```
marcan: Date needs updating to 2017.

media-plugins/distrho-ports/metadata.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
+    <email>webmaster@gentoostudio.org</email>
```
marcan: This could use a `<name>` element ("Gentoo Studio Project" or similar? whatever you feel is appropriate here)

media-plugins/distrho-ports/distrho-ports-9999.ebuild
marcan: Also the project website mentions more deps:

> ALSA
> freetype2
> OpenGL/Mesa
> X11 core and extensions (XShm, XRender and XCursor)

Are these really not required? Or are they optional? If they're optional they should be USE flags to avoid automagic dependencies.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+DESCRIPTION="Linux ports of Distrho plugins."
+HOMEPAGE="http://distrho.sourceforge.net"
+
+SRC_URI=""
+EGIT_BRANCH="master"
+EGIT_REPO_URI="https://github.com/DISTRHO/DISTRHO-Ports.git"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
```
marcan: Since this is a live ebuild, it should probably have to keywords so it has to be explicitly unmasked. However, since this project seems to have no releases, maybe it's okay to have this as ~amd64 for now. Not sure, just commenting :)

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: No, it doesn't.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: I thought of updating this, so I will do so.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Thanks, I was wondering why I see a lot of undef SRC_URI's.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: ALSA and Freetype I can add, but since these can't be used without a DE or WM already installed, X and opengl should already exist.

media-plugins/distrho-ports/metadata.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
```
simonvanderveldt: Can you switch this to a person (you) just like the other packages?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Added alsa-lib and changed premake to premake:3. (Will upload/fix.)

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+src_prepare() {
+	"${WORKDIR}"/"${P}"/scripts/premake-update.sh linux
+}
+
+src_compile() {
+	emake
+}
+
+src_install() {
+	emake DESTDIR="${D}" install
```
simonvanderveldt: I don't think this is necessary, should be covered by the default, right?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
simonvanderveldt: AFAIK 9999 ebuilds should never have a KEYWORD.
We can requests a release though

media-plugins/distrho-ports/distrho-ports-9999.ebuild
simonvanderveldt: `repoman full` should've shown this, did you run it?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Good idea, will fix that.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Will give this a try.

media-plugins/distrho-ports/metadata.xml
audiodef: Done, will fix/upload.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: I know less about this than you, so I'll go with whatever you decide.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Marcan suggested removing it, so I'm about to try that.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Failed for me without specifying workdir, so I'll leave that alone.

audiodef:
Make changes, waiting for a recompile on my VM to make sure it goes smoothly here, then will copy changes to PR.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Went ahead and took the keyword out. You're right, live ebuilds don't have keywords.

media-plugins/distrho-ports/metadata.xml
audiodef: I think "project" is appropriate here, which is what Gentoo Studio is. A lot of what I implement is because of community input on FGO. (I thought I commented on this, but it seems to have disappeared.)

domichel: I agree with audiodef.

simonvanderveldt: Hmm, I'd very much prefer this to be a person. First of all because then we can easily see who the maintainer is. Second of all because I'd like to add a feature that assigns issues relating to packages to the maintainer here on github and we need the info for that.
Also, if we'd add a project here it should be "gentoo-audio", but because of the above IMHO we shouldn't.

Regarding input from others, if there are people who want to contribute please point them to this repo so they can either create an issue here with a description of what they want or, even better, create a PR to implement their fixes or join the team if they want to maintain a package.

simonvanderveldt:
Looking at the error message from the emerge
```
In file included from ../source/modules/juce_audio_processors/juce_audio_processors.cpp:168:0:
../source/modules/juce_audio_processors/format_types/juce_LADSPAPluginFormat.cpp:29:20: fatal error: ladspa.h: No such file or directory
 #include "ladspa.h"
```

It seems like you're missing a dependency on `media-libs/ladspa-sdk`

media-plugins/distrho-ports/metadata.xml
audiodef: Thanks for explaining. I will make this change.

simonvanderveldt:
@audiodef Do you maybe want to "tag"/use a specific commit that you know works, similar to what @marcan did for #48 ?

media-plugins/distrho-ports/metadata.xml
simonvanderveldt: Thx! :)

media-plugins/distrho-ports/distrho-ports-9999.ebuild
simonvanderveldt: @audiodef These deps should all be defined. In case you're not sure it usually helps to search through the ebuilds in the gentoo tree
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=freetype&type=&utf8=%E2%9C%93
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=xrender&type=&utf8=%E2%9C%93
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=xcursor&type=&utf8=%E2%9C%93

As @marcan said if they're optional they should be behind a USE flag to avoid automagic dependency usage.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,26 @@
...
+DESCRIPTION="Linux ports of Distrho plugins."
+HOMEPAGE="http://distrho.sourceforge.net"
+
+EGIT_BRANCH="master"
+EGIT_REPO_URI="https://github.com/DISTRHO/DISTRHO-Ports.git"
+
+LICENSE="GPL-2"
+SLOT="0"
+
+RDEPEND="media-libs/alsa-lib
```
simonvanderveldt: Seems like you're missing some deps here as @marcan mentioned https://github.com/gentoo-audio/audio-overlay/pull/50#discussion_r124906114

These deps should all be defined. In case you're not sure I find it helps to search through the ebuilds in the gentoo tree for how it's done there:
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=freetype&type=&utf8=%E2%9C%93
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=xrender&type=&utf8=%E2%9C%93
- https://github.com/gentoo/gentoo/search?l=Gentoo+Ebuild&q=xcursor&type=&utf8=%E2%9C%93

As @marcan said if they're optional they should be behind a USE flag to avoid automagic dependency usage.

audiodef: I'll make some changes.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,29 @@
...
+EGIT_BRANCH="master"
+EGIT_REPO_URI="https://github.com/DISTRHO/DISTRHO-Ports.git"
+
+LICENSE="GPL-2"
+SLOT="0"
+
+IUSE="+alsa +jack"
+
+RDEPEND="alsa? ( media-libs/alsa-lib )
+        jack? ( virtual/jack )"
```
mrVanDalo: I don't think these dependencies are enough.
When  I look at the README.md I see at least `OpenGL/Mesa` and `X11` which is missing.

You can use 
```
qlist -e distroh-ports |  grep -ve "'" | xargs scanelf -L -n -q -F "%n #F" |  tr , ' ' | xargs qfile -Cv | sort -u | awk '{print $1}' | uniq | xargs qatom --format  "%{CATEGORY}/%{PN}" | xargs -L1 qdepends --nocolor --name-only --rdepend --query | grep ^virtual  | uniq
```
to find out the `RDEPEND` packages.

simonvanderveldt:
> Please merge the commits. https://wiki.gentoo.org/wiki/Gentoo_GitHub#Rebasing_commits

@mrVanDalo Whilst I prefer this as well it's not necessary given the way we currently work. GitHub will do that for us if we use "Squash and merge" :)

mrVanDalo:
@simonvanderveldt true, github can do this. But you can quickly forget this. Also sometimes a pull-request should have multiple commits ( https://wiki.gentoo.org/wiki/Gentoo_GitHub#Separate_.28or_.22atomic.22.29_commits ). It is just confusing seeing a lot of commits, and I will always point that out, because this is something which is explicitly documented (and very reasonable).

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: I deliberately left these out because it doesn't make sense to me that anyone would even bother installing this kind of software unless they already have a desktop environment already set up.

audiodef: Oh, I see why you mentioned it. Travis:

../source/modules/juce_events/juce_events.cpp:58:23: fatal error: X11/Xlib.h: No such file or directory

It won't hurt to include it, so I will.

mrVanDalo: Sorry I don't understand. You left it out because somebody who what's to install this package already has all dependencies installed?

simonvanderveldt:
Nice, the build passed! :)

Couple of observations/issues from the build output:
- There are a lot of pre-stripped files, can you file a bug to upstream to fix this? See for an example how to fix this https://github.com/falkTX/Carla/pull/475
And please add the `QA_PRESTRIPPED` variable to the ebuild with all the prestripped binaries in it.
- A lot of files are being installed to `/usr/src/`. That shouldn't be the case
- The plugins are installed to `/usr/local/lib/`, this should be `/usr/lib`

And one question, is it possible to choose which formats are being built? I see both lv2 and vst binaries as part of the build.

> Added x11-libs/libx11 to build deps, doesn't seem like a runtime dep.

Are you sure about this? I'd expect since at least some of the LV2 plugins show a UI that they will probably need it runtime?

media-plugins/distrho-ports/distrho-ports-9999.ebuild
audiodef: Nevermind, I'm the one being slow. X11 deps added. :)

audiodef:
> There are a lot of pre-stripped files, can you file a bug to upstream to fix this? See for an example how to fix this falkTX/Carla#475

Feel free to file this bug. I would not know what I'm talking about here. 

> And please add the QA_PRESTRIPPED variable to the ebuild with all the prestripped binaries in it.

This is completely new to me. Could you add this?

> A lot of files are being installed to /usr/src/. That shouldn't be the case

Definitely not! How should this be fixed?

> The plugins are installed to /usr/local/lib/, this should be /usr/lib

I agree. I can try to address this.

> And one question, is it possible to choose which formats are being built? I see both lv2 and vst binaries as part of the build.

This is a good idea. I prefer LV2 over VST, so I would exercise a set of +lv2 -vst use flags. 

> Added x11-libs/libx11 to build deps, doesn't seem like a runtime dep.
> Are you sure about this? I'd expect since at least some of the LV2 plugins show a UI that they will probably need it runtime?

I will defer to you. It seemed like it wouldn't even build without these deps, so I dropped them into DEPEND instead of RDEPEND.

audiodef:
Added S="${WORKDIR}", which seems to take care of some of the issues mentioned regarding install locations.

simonvanderveldt:
> Added S="${WORKDIR}", which seems to take care of some of the issues mentioned regarding install locations.

@audiodef Now it doesn't compile anything at all ;)
I don't think `S` is the issue here, it was picking up the sources fine. Just needs to honor `DESTDIR="${D}"` and find a way to not install all the source files.

audiodef:
I thought emerge was a little too quick after declaring S. Ok, back to drawing board...

media-plugins/distrho-ports/distrho-ports-9999.ebuild
marcan: Just to clarify, all dependencies are always necessary. It doesn't matter that most people would already have those deps installed. Not having them breaks package management in subtle ways (and of course also breaks things like Travis and other automated test systems which *don't* have those deps installed). Anything the build system for a package touches on the host system at all must be a dependency, except for stuff in @system which I think you're allowed to assume is there (at least tools-wise; libraries other than glibc I'd still include as deps, especially if sub-slots are involved).

marcan:
@audiodef are you sure libx11 is a build dep only? That doesn't make sense; most libraries are runtime deps unless they are only used by the build system (and there's no reason why a build system would use libx11).

audiodef:
I'm not all that sure, so I will defer to your knowledge about this. Will go change the ebuild accordingly.

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+EGIT_BRANCH="master"
+EGIT_REPO_URI="https://github.com/DISTRHO/DISTRHO-Ports.git"
+
+LICENSE="GPL-2"
+SLOT="0"
+
+IUSE="+alsa +jack"
+
+RDEPEND="alsa? ( media-libs/alsa-lib )
+        jack? ( virtual/jack )
```
mrVanDalo: I can't see jack in the dependencies in the [README](https://github.com/DISTRHO/DISTRHO-Ports) (but this means nothing of course :D) 
But It would be nice if you could add lv2 support  ( `make lv2` has to be called for that )

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+	x11-libs/libX11
+	x11-libs/libXext
```
mrVanDalo: I still cant see the `virtual/opengl` or `media-libs/mesa` which is demanded on the [README](https://github.com/DISTRHO/DISTRHO-Ports)

media-plugins/distrho-ports/distrho-ports-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+EGIT_BRANCH="master"
+EGIT_REPO_URI="https://github.com/DISTRHO/DISTRHO-Ports.git"
+
+LICENSE="GPL-2"
+SLOT="0"
+
+IUSE="+alsa +jack"
+
+RDEPEND="alsa? ( media-libs/alsa-lib )
```
mrVanDalo: These are the runtime dependencies which I found : 
```media-libs/alsa-lib
media-libs/freetype
x11-libs/libX11
x11-libs/libXext
```

mrVanDalo: But I did not check if the programs work.

mrVanDalo:
The travis build just installes git and than fails.
@simonvanderveldt did you see that ?

simonvanderveldt:
> The travis build just installes git and than fails.
> @simonvanderveldt did you see that ?

@mrVanDalo No, it emerged the ebuild and it's dependencies correctly, see https://travis-ci.org/gentoo-audio/audio-overlay/jobs/255027674

So the emerge is working again, which is nice :)
But we're now pretty much back to this point https://github.com/gentoo-audio/audio-overlay/pull/50#issuecomment-313205947
Apart from the files in `/usr/src`, these have been removed.

simonvanderveldt:
@audiodef I wanted to use obxd so spent some time fixing this ebuild. Hope you don't mind :)

I've rebased this on current `master` and added a new commit that fixes the remaining issue. I've used the LV2 version of some of the plugins and they work fine and there's no cruft being installed anymore.

Let's see if the tests pass and then get this merged once everyone is happy with it :)

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -0,0 +1,83 @@
...
+LICENSE="GPL-2"
+SLOT="0"
+
+IUSE="lv2 vst"
+REQUIRED_USE="|| ( lv2 vst )"
+
+RDEPEND="media-libs/alsa-lib
+  media-libs/freetype
+  media-libs/ladspa-sdk
+  media-libs/mesa
```
mrVanDalo: I think this is covered by `virtual/opengl`

simonvanderveldt: You're right :)
I've updated the ebuilds to use `virtual/opengl` instead of `media-libs/mesa`

PR #51, opened 2017-06-30
Add media-sound/ladish

### LINKED ISSUE #45: Add ladish

simonvanderveldt:
Fixes #45


trebmuh:
Hi guys,
I don't know if you're aware of this repo : https://github.com/alessio/ladish . IIRC this is the one which had become the "upstream" one. At least on debian, it is. Need to be double check tho.

simonvanderveldt:
> Hi guys,
> I don't know if you're aware of this repo : https://github.com/alessio/ladish . IIRC this is the one which had become the "upstream" one. At least on debian, it is. Need to be double check tho.

@trebmuh Yeah, I know :) Thanks for the pointer though. The only new commit in there is actually mine ;)
I'm in touch with nedko to see if these things can be merged into ladish upstream. Sounds like it's doable, but we'll see :)

trebmuh:
Great.

Last time I've discussed it with nedko on IRC, he told me that he thought that alessio had access to nedko's repo (which isn't the case IIRC).

It might be a good time to ping @nedko and @alessio about this, then only one repo would be existing, facilitating everything for users and packagers I guess.

I don't want to push anyone here so, please guys, just take this as a gently reminder and feel free to throw it to the window if this comment not useful enough :)

media-sound/ladish/ladish-1.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+EAPI=6
+
+PYTHON_COMPAT=( python2_7 )
+PYTHON_REQ_USE='threads(+)'
+
+inherit flag-o-matic python-single-r1 waf-utils
+
+DESCRIPTION="LADI Session Handler - a session management system for JACK applications"
+HOMEPAGE="http://ladish.org/"
```
mrVanDalo: This page says : 
`ladish is not dead, the site is down`

media-sound/ladish/ladish-1.ebuild
```diff
@@ -0,0 +1,79 @@
...
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/LADI/${PN}.git"
+	KEYWORDS=""
+else
+	inherit vcs-snapshot
```
mrVanDalo: what do you need this `eclass` for ?

media-sound/ladish/ladish-1.ebuild
```diff
@@ -0,0 +1,79 @@
...
+EAPI=6
+
+PYTHON_COMPAT=( python2_7 )
+PYTHON_REQ_USE='threads(+)'
+
+inherit flag-o-matic python-single-r1 waf-utils
+
+DESCRIPTION="LADI Session Handler - a session management system for JACK applications"
+HOMEPAGE="http://ladish.org/"
+if [[ ${PV} == *9999 ]]; then
```
mrVanDalo: I still find this confusing. But it's Ok

media-sound/ladish/ladish-1.ebuild
simonvanderveldt: Good point, I guess I'll update it to https://github.com/LADI/ladish for now

media-sound/ladish/ladish-1.ebuild
simonvanderveldt: Well spotted! :)
Probably don't need it, I'll check and fix if that's the case

simonvanderveldt:
> Is somebody using this? the project looks pretty dead.

I use it all the time when I do "modular" stuff :)
Still works the best of all the session management tools imho.

As @trebmuh pointed out some activity has moved to a fork/forks, but I expect those changes to be merged eventually.

PR #54, opened 2017-07-01
Tests: Fix gitr3 emerge test failure

### LINKED ISSUE #53: Emerge test fails for git-r3 ebuild with only binary packages as dependencies
If we have all the dependencies available as binary packages portage doesn't need to check out any source files for any dependencies, which means `/usr/portage/distfiles` isn't created.
The first time portage needs to checkout source files is when the 9999 ebuild is being built. And apparently that doesn't work because it's not allowed to create the `/usr/portage/distfiles` directory because `/usr/portage` is owned by root:root and the process runs as user `portage`

This is a know bug/issue with git-r3 related to https://bugs.gentoo.org/481434

I'm looking at some workarunds, will create a PR asap.

simonvanderveldt:
by using /tmp/distfiles as the portage distfiles dir. This directory is always writable by every user so doesn't fail.

Fixes #53

### DIFF
--- a/tests/emerge_ebuild.sh
+++ b/tests/emerge_ebuild.sh
@@ -18,6 +18,13 @@ export FEATURES="-news" PORTAGE_RSYNC_EXTRA_OPTS="-q"
 # Update the portage tree
 emerge --sync
 
+# Set portage's distdir to /tmp/distfiles
+# This is a workaround for a bug in portage/git-r3 where git-r3 can't
+# create the distfiles/git-r3 directory when no other distfiles have been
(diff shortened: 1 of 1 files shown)


domichel:
Most users will not need that export. Which imply I will appreciate if it is some test like
```
if [ ! -e /usr/portage/distfiles ]
    then export DISTDIR="/tmp/distfiles"
fi

```

simonvanderveldt:
@domichel this is not meant to be run locally but in the container started by `emerge_new_or_changed_ebuilds.sh`.

PR #58, opened 2017-07-05
Add media-sound/pure-data

audiodef:


media-sound/pd/Manifest
```diff
@@ -0,0 +1,4 @@
+DIST pd-0.47-1.src.tar.gz 2296697 SHA256 c728e694bdb17af0836b084652938e4b8c2f57401fd5014fe2fe343c1349ba4c SHA512 29c690ca1746f8a9beb353f58b584a1ea298a6dbbad951111a91cb14900a0a7913e790806c2d8280a2099d8acb6e693b2a52555821ab92733d125fcfd5795b9c WHIRLPOOL dd0d1fdb02361acae6d50300d7aa29890726ed62307be4ae4dfbbacc334cf459c06809fb7aedfb1b16d02e29c262a2951aa939bd987c2fa2813ffac90e938205
+DIST pd.png 1048 SHA256 991c5bbe3b0e4f3242064e56de914d853ca7cc6b339b5763023dad83dc9854b7 SHA512 bbb8f33a041ea8edf62b0e0d9adbf95515e78b8dae0ee5407fe30092282fd2e614713248b5c86d8e0974f8abb30e5c1089952ad947963009f026024f81fd2c9c WHIRLPOOL 7076a677b0eb567aa1c1a4a0819578a375b380c3cb8c052430b45231556ff79d975844e25930798d7fdc80c3d78757c0d23ef91f7f1c1d53f5343aa5d082c4ba
+EBUILD pd-0.47_p1.ebuild 1123 SHA256 4ae97d7da62d4952ef1c6d4bf81a0b2be9776642b3b5710fbc042c43abfb96b2 SHA512 9a76ac8637e5b03adbeaf8a8d24dd9e6339983416809ba4ac11ef829b9892ee090d4195dee06f5221736af620a8a9bb529637512d4869d33d54dba8a68464098 WHIRLPOOL f8ea8d85f23c0ea4ff1b2956e82222e9d5c4616fb753ddf48efe464bc70a7b783be37823c19892a7938581761af567bdebf4cab8ad8fbf47d68a484fc808e0ba
```
simonvanderveldt: @audiodef This shouldn't exist here since we have thin manifests enabled. Same as the line below.
How did you generate this file?

simonvanderveldt:
There's an issue with the current release (47-1) and master of pure data that currently prevent it from working without patches when trying to use jack.

The initial issue is that pure data isn't being linked to jack because of the inclusion of `weakjack.h` which causes jack to be filtered out of the libs to be linked to.

Once the linking works I end up with a segfault in jack, see https://github.com/pure-data/pure-data/pull/93#issuecomment-313871401 for the backtrace of it.

Both of these issues are solved for me if `weakjack.h` isn't included. See https://github.com/pure-data/pure-data/pull/93 which now includes this fix and [this pure data mailing list thread/message](https://lists.puredata.info/pipermail/pd-dev/2017-07/021099.html).

I don't know if there's a better way to fix this issue, but depending on if that PR gets merged or not I think we'll to provide a patch to fix this.

simonvanderveldt:
OK, the pure data PR to fix the jack issues has been merged. Meaning for the 9999 ebuild it should be fine, but for the 47-1 ebuild we'll have to provide a patch for this commit https://github.com/pure-data/pure-data/commit/17b9ec6dd141eb8eeb7c17d083ebe42fc2c43605
I'll update this branch/PR so it's included.

simonvanderveldt:
OK, pure-data-0.47_p1 now works for me, with both ALSA and JACK.
For some reason pure-data-9999/from `master` segfaults though when connecting to JACK :|
I'll file a bug upstream and see if we can figure out why it's happening.

media-sound/pure-data/pure-data-0.47_p1.ebuild
```diff
@@ -0,0 +1,69 @@
...
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/pure-data/${PN}.git"
+	KEYWORDS=""
+else
+	MY_PN="pd"
+	MY_PV=${PV/_p/-}
+	MY_P="${MY_PN}-${MY_PV}"
+	SRC_URI="http://msp.ucsd.edu/Software/${MY_P}.src.tar.gz
+					 https://puredata.info/portal_css/Plone%20Default/logo.png -> ${PN}.png"
```
mrVanDalo: strange indention

simonvanderveldt: Yeah, that's because ebuilds have to be indented with tabs, so it depends on the editor/viewer's settings for tab-width.
Maybe it also works without the extra indentation, let me try.

simonvanderveldt:
> For some reason pure-data-9999/from master segfaults though when connecting to JACK :|
> I'll file a bug upstream and see if we can figure out why it's happening.

I think for now we should skip including the 9999 ebuild, once someone has figured out why it isn't working we can add it.

simonvanderveldt:
@audiodef confirmed that 0.47_p1 is working for him as well when using JACK, so we can merge this one :)

PR #67, opened 2017-07-16
Add media-sound/sequencer64

### LINKED ISSUE #55: Add sequencer64
The in active development version/fork of seq24 https://github.com/ahlstromcj/sequencer64
I already have an ebuild, will do some cleanup and create a PR

simonvanderveldt:
Fixes #55


simonvanderveldt:
CI fails because of a missing alsa dependency. I've opened ahlstromcj/sequencer64#98 to check if this is a hard or optional dependency

simonvanderveldt:
I'm having some issue with seq64 not showing up as a JACK client. Didn't happen with previous versions, but might also be caused by something else. I'll troubleshoot and update here once I know more.

audiodef:
Installed 9999 on my Gentoo Studio dev vm. It showed up as a JACK client with a MIDI in on Carla/Catia.

simonvanderveldt:
@audiodef thanks for checking and the confirmation it works for you.
Do you use lash/ladish by accident? Somehow it seems it's not working in the lash/ladish context, if I add seq64 to a ladish room it doesn't shop up at all.

I can reproduce my issue now, though I haven't figured out why it's happening.
Will probably have some time tomorrow to see if I can find the cause.

audiodef:
@simonvanderveldt On Gentoo Studio, both lash and ladish get pulled in, but I don't personally use them.

simonvanderveldt:
OK, I found the commit that's causing the issue and created ahlstromcj/sequencer64#100 for it.
I'll include the 0.91.0 version in this PR as well, since that's the last released version that works.

simonvanderveldt:
@ahlstromcj has [commited a fix](https://github.com/ahlstromcj/sequencer64/commit/fd5567684ffb4fc9926f7b57befa84675b470ad7) for the above issue. I've included that commit as a patch so everything works as expected again for both 0.93.0 as well as current master.
Because of this there's no longer the need to add the older 0.91.0 version.

This was the final item to fix, so I'm gonna go ahead and merge this :)

PR #70, opened 2017-07-27
Add media-sound/helm

### LINKED ISSUE #16: add helm
http://tytel.org/helm/

simonvanderveldt:
Fixes #16 

For some reason the standalone/jack version isn't really working for me, as in I don't have any JACK MIDI in ports :x
But this is also already the case with the older versions, seems like something is going wrong when compiling since the packages from for example KX Studio do work afaik.
See https://github.com/mtytel/helm/issues/152

Can someone check if this also happens for you?


audiodef:
I will try this out as soon as I get a chance. Been sick lately.

audiodef:
I can confirm that this happens to me, too. There are no JACK inputs, even with a2j. The only way I can play the Helm is to use my laptop's keyboard - and that's no fun.

audiodef:
Could there be a jack compile option that needs the jack use flag?

simonvanderveldt:
> Could there be a jack compile option that needs the jack use flag?

@audiodef  Thats what I expected, but I haven't been able to find one https://github.com/mtytel/helm/blob/master/Makefile
Maybe it's a JUCE option/issue?

I think the simplest option for now is to remove the `standalone` USE flag/options, so one can build the lv2 and/or vst plugins, since those do work.

audiodef:
I know that Juce gets pulled in for some things, but I'm not familiar with how it works.

I vote we leave the standalone use flag there, and ewarn that there are no jack inputs. It is functional via computer keyboard, and some people might not want to use an LV2.

media-sound/helm/helm-9999.ebuild
```diff
@@ -0,0 +1,58 @@
...
+	EGIT_REPO_URI="https://github.com/mtytel/helm.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://github.com/mtytel/helm/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+SLOT="0"
+
+IUSE="standalone lv2 vst"
+REQUIRED_USE="|| ( standalone lv2 vst )"
```
mrVanDalo: `vst` ? this thing can compile without any vstsdk ? (as I can see you don't user travis to test for vst build)

tests/packages/media-sound/helm-9999.conf
```diff
@@ -0,0 +1,3 @@
+echo "media-sound/helm lv2 standalone" >> /etc/portage/package.use/helm
```
mrVanDalo: a bit simpler : 
```shell
cat >/etc/portage/package.use/helm <<EOF
media-sound/helm lv2 standalone
media-libs/mesa -llvm
EOF

echo "VIDEO_CARDS=intel" >> /etc/portage/make.conf
```

media-sound/helm/helm-9999.ebuild
simonvanderveldt: Yeah, since the VST3 SDK was made available under a usable license some projects have included it in their codebase. Helm does that as well, see https://github.com/mtytel/helm/tree/master/VST3_SDK

mrVanDalo: Yeah that is Ok, i guess. But why is VST not a "sane default". I like to test as much as possible, for a small price of course. If you have to make a lot of special adjustments to test a small edge-case, that's not a good relation between workload and benefit. But here it looks like you just have to add 3 letters.

simonvanderveldt: Mainly because vst used to be problematic because of the SDK dependency.
Also, it's an opinionated choice since we have LV2 on Linux I don't see much reason to use/build VST.

But you're right, it probably makes more sense to test as much as possible and since the VST SDK is included in the sources I don't expect it to cause a lot of issue, so I'll add it to the test USE config as well.

simonvanderveldt:
> I vote we leave the standalone use flag there, and ewarn that there are no jack inputs. It is functional via computer keyboard, and some people might not want to use an LV2.

I'm starting to get the feeling that JUCE simply doesn't support JACK MIDI and afaik Helm uses JUCE as it's "backend". So I guess it's to be expected that there's no JACK MIDI input.
Do you still think an ewarn is necessary?

> What about contacting the KXStudio guy to ask how he got JACK inputs working? In addition to the KXStudio website, I've seen him on the Ardour forums.

"The KXStudio guy" is @falktx ;)
I've pinged him on IRC to check if he ran into the same issue.

falkTX:
Juce based plugins as standalone do not have JACK MIDI support.
In juce+linux, MIDI always goes through ALSA.

simonvanderveldt:
@falkTX Thanks for the confirmation!

I'll see if I can get ALSA to work.

simonvanderveldt:
Closing this, I'm working on getting helm bumped upstream, see https://bugs.gentoo.org/638150

PR #71, opened 2017-07-28
fix travis build for bitwig and praxis-live

### LINKED ISSUE #65: Fix package specific emerge tests
Some packages currently fail the emerge test, we should fix that :)

List of packages:
- praxis-live-bin
- bitwig-studio

mrVanDalo:
fixes #65

### DIFF
--- a/tests/packages/media-sound/bitwig-studio-1.3.16.conf
+++ b/tests/packages/media-sound/bitwig-studio-1.3.16.conf
@@ -1,0 +1,1 @@
+bitwig-studio.conf
--- a/tests/packages/media-sound/bitwig-studio-2.1.1.conf
+++ b/tests/packages/media-sound/bitwig-studio-2.1.1.conf
@@ -1,0 +1,1 @@
+bitwig-studio.conf
--- a/tests/packages/media-sound/bitwig-studio.conf
+++ b/tests/packages/media-sound/bitwig-studio.conf
@@ -1,0 +1,3 @@
+cat >/etc/portage/package.use/bitwig-studio <<EOF
+media-video/ffmpeg cpu_flags_x86_mmxext
+EOF
--- a/tests/packages/media-video/praxis-live-bin-3.1.0.conf
+++ b/tests/packages/media-video/praxis-live-bin-3.1.0.conf
@@ -1,0 +1,1 @@
+praxis-live-bin.conf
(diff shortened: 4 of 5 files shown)


mrVanDalo:
@simonvanderveldt sorry I don't get the error message from travis. It seems there is something wrong with the testing script.

mrVanDalo:
Sorry don't get the error message.

simonvanderveldt:
@mrVanDalo The config files for the emerge test should be called `<P>.conf`, right now they're called `<P>.ebuild`.

Note that currently changes to the `.conf` files won't trigger a test for the matching package.

mrVanDalo:
@simonvanderveldt Ok so how do we find out if the changes work?

simonvanderveldt:
@mrVanDalo You can run the tests locally if you want?

Run from within the repo

```
docker run --rm -ti -v ~/.portage-pkgdir:/usr/portage/packages -v $PWD:/usr/local/portage -w /usr/local/portage gentoo/stage3-amd64:latest /usr/local/portage/tests/emerge_ebuild.sh <ebuild-path>
```

For example:
```
docker run --rm -ti -v ~/.portage-pkgdir:/usr/portage/packages -v $PWD:/usr/local/portage -w /usr/local/portage gentoo/stage3-amd64:latest /usr/local/portage/tests/emerge_ebuild.sh media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
```
Note that the volume mount for portage-pkgdir isn't strictly necessary, but it will save a lot of time if you run the tests more than once.

I want to update the test scripts so we can run them straight from the repo, but there will be some conventions in there (like the pkgdir volume) that might not be how everyone wants to run them.

simonvanderveldt:
@mrVanDalo I'm considering just setting/adding `mmxext` to `CPU_FLAGS_X86` in `make.conf`
That could be a generic thing we do for every test since `mmxext` should actually if/when `sse` is enabled. That way we don't have to do this fix for every package that depends on ffmpeg (or any other package that for some reason needs `mmxext`.

What do you think?

mrVanDalo:
Sounds also good. I like configuration stuff in one place only. so generating on top complexity (another place to look on how stuff is set up) is not in my favor. But yeah changing the make.conf for theses builds also sounds good.

mrVanDalo:
@simonvanderveldt See #76 Seems your idea works. When you approved this pr I'll create a fix for praxis-live :D

simonvanderveldt:
This has been fixed by #111 and #126

PR #72, opened 2017-07-31
tests/refactor : try to stick to 80 letters per line

mrVanDalo:
I try to make scripts more readable : 

* use long options of shell commands (more descriptive)
* one option per line.

### DIFF
--- a/tests/emerge_ebuild.sh
+++ b/tests/emerge_ebuild.sh
@@ -13,7 +13,8 @@ EBUILD="${1}"
 echo "Emerging ${EBUILD}"
 
 # Disable news messages from portage and disable rsync's output
-export FEATURES="-news" PORTAGE_RSYNC_EXTRA_OPTS="-q"
+export FEATURES="-news" \
+       PORTAGE_RSYNC_EXTRA_OPTS="-q"
 
 # Update the portage tree
 emerge --sync
@@ -43,7 +44,16 @@ if [ -f "${PKG_CONF_FILE}" ]; then
 fi
 
 # Emerge dependencies first
-emerge --quiet-build --buildpkg --usepkg --onlydeps --autounmask=y --autounmask-continue=y "=${PKG_CATEGORY}/${PKG_NAME}"
+emerge \
+    --quiet-build \
+    --buildpkg \
+    --usepkg \
+    --onlydeps \
+    --autounmask=y \
+    --autounmask-continue=y \
+    "=${PKG_CATEGORY}/${PKG_NAME}"
 
 # Emerge the ebuild itself
-emerge -v "=${PKG_CATEGORY}/${PKG_NAME}"
+emerge \
+    --verbose \
+    "=${PKG_CATEGORY}/${PKG_NAME}"
--- a/tests/emerge_new_or_changed_ebuilds.sh
+++ b/tests/emerge_new_or_changed_ebuilds.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
+#
 # Determine which ebuilds are new or changed and emerge them
+
 set -ex
 
 # Get list of new or changed ebuilds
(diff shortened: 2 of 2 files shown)


tests/emerge_ebuild.sh
```diff
@@ -13,7 +13,8 @@ EBUILD="${1}"
 echo "Emerging ${EBUILD}"
 
 # Disable news messages from portage and disable rsync's output
-export FEATURES="-news" PORTAGE_RSYNC_EXTRA_OPTS="-q"
+export FEATURES="-news" \
```
simonvanderveldt: this should be a single line

tests/emerge_ebuild.sh
```diff
@@ -43,7 +44,16 @@ if [ -f "${PKG_CONF_FILE}" ]; then
...
+    --buildpkg \
+    --usepkg \
+    --onlydeps \
+    --autounmask=y \
+    --autounmask-continue=y \
+    "=${PKG_CATEGORY}/${PKG_NAME}"
 
 # Emerge the ebuild itself
-emerge -v "=${PKG_CATEGORY}/${PKG_NAME}"
+emerge \
```
simonvanderveldt: this should be a single line?

tests/emerge_new_or_changed_ebuilds.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
+#
```
simonvanderveldt: why this superfluous whitespace?

tests/emerge_new_or_changed_ebuilds.sh
```diff
@@ -1,5 +1,7 @@
...
 # Determine which ebuilds are new or changed and emerge them
+
```
simonvanderveldt: why this superfluous whitespace?

tests/emerge_ebuild.sh
mrVanDalo: for bash it is but for us humans we it is in multiple lines so we can read it better.

tests/emerge_new_or_changed_ebuilds.sh
mrVanDalo: It is there because of readability.

tests/emerge_new_or_changed_ebuilds.sh
mrVanDalo: Same here. It separates the script header from the script.

mrVanDalo:
> Don't you think the single argument per line is a bit excessive?

I think : "Code is there to be read by other programmers".
Do you think it is worse to read, or harder to write?

I can definitively read this 
```
emerge \
    --quiet-build \
    --buildpkg \
    --usepkg \
    --onlydeps \
    --autounmask=y \
    --autounmask-continue=y \
    "=${PKG_CATEGORY}/${PKG_NAME}"
```
better than this 
```
emerge --quiet-build --buildpkg --usepkg --onlydeps --autounmask=y --autounmask-continue=y "=${PKG_CATEGORY}/${PKG_NAME}"
```
And sometimes it is not necessary to break the line to achieve the 80 chars per line rule. But I do it anyway because it improves the readability.

mrVanDalo:
> The long form arguments I can understand even though I find them to be too verbose most of the time.

well it is good they are verbose. It helps to read the command.

mrVanDalo:
> 80 chars per line is fine as a basic default and a good idea to fix

I don't want to make this a rule. It is just a personal rule to me. I don't know why I wrote it, sorry this should not be part of the discussion.

tests/emerge_new_or_changed_ebuilds.sh
simonvanderveldt: OK, I'd like to suggest to use google's shell script style/formatting guide for our shell scripts so we can ensure they are consistent.
See https://google.github.io/styleguide/shell.xml

In there they have the same format for the file header comment :)

simonvanderveldt:
> I think : "Code is there to be read by other programmers".
> Do you think it is worse to read, or harder to write?
> 
> I can definitively read this
> 
> emerge \
>     --quiet-build \
>     --buildpkg \
>     --usepkg \
>     --onlydeps \
>     --autounmask=y \
>     --autounmask-continue=y \
>     "=${PKG_CATEGORY}/${PKG_NAME}"
> better than this
> 
> emerge --quiet-build --buildpkg --usepkg --onlydeps --autounmask=y --autounmask-continue=y "=${PKG_CATEGORY}/${PKG_NAME}"
> And sometimes it is not necessary to break the line to achieve the 80 chars per line rule. But I do it anyway because it improves the readability.

After looking at it a bit more, the split one does seem more legible, thanks for the example!

So let's use this style for lines that are longer than 80 characters (same as in [google's styleguide](https://google.github.io/styleguide/shell.xml?showone=Line_Length_and_Long_Strings#Line_Length_and_Long_Strings))

simonvanderveldt:
The majority of the tests have been updated so they don't really have long lines anymore, closing this one

PR #76, opened 2017-08-13
media-sound/bitwig-studio: fix jack2 dependencies

mrVanDalo:
I tested it, bitwig needs API calls which only exist in jack2

### DIFF
--- a/tests/packages/media-sound/bitwig-studio-1.3.16.conf
+++ b/tests/packages/media-sound/bitwig-studio-1.3.16.conf
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+cat >/etc/portage/package.use/bitwig-studio <<EOF
+media-libs/mesa -llvm
+EOF
+
+cat >>/etc/portage/make.conf <<EOF
+VIDEO_CARDS="intel"
+CPU_FLAGS_X86="sse mmxext mmx"
+EOF
--- a/media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
+++ b/media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild
@@ -23,7 +23,7 @@ RDEPEND="${DEPEND}
 	app-arch/bzip2
 	dev-libs/expat
 	gnome-extra/zenity
-	jack? ( virtual/jack )
+	jack? ( media-sound/jack2 )
 	media-libs/alsa-lib
 	media-libs/fontconfig
 	media-libs/freetype
--- a/media-sound/bitwig-studio/bitwig-studio-2.1.1.ebuild
+++ b/media-sound/bitwig-studio/bitwig-studio-2.1.1.ebuild
@@ -23,7 +23,7 @@ RDEPEND="${DEPEND}
 	app-arch/bzip2
 	dev-libs/expat
 	gnome-extra/zenity
-	jack? ( virtual/jack )
+	jack? ( media-sound/jack2 )
 	media-libs/alsa-lib
 	media-libs/fontconfig
 	media-libs/freetype
--- a/tests/packages/media-sound/bitwig-studio-2.1.1.conf
+++ b/tests/packages/media-sound/bitwig-studio-2.1.1.conf
@@ -1,0 +1,1 @@
+bitwig-studio-1.3.16.conf


mrVanDalo:
@simonvanderveldt this thing fails because it exceeds time. I'll try to shorten it tomorrow, but I don't think it will workout.

tests/packages/media-sound/bitwig-studio-1.3.16.conf
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+cat >/etc/portage/package.use/bitwig-studio <<EOF
```
simonvanderveldt: Can you change this into a simple `echo "media-libs/mesa -llvm" > /etc/portage/package.use/bitwig-studio`?

simonvanderveldt:
This fixes part of #65

simonvanderveldt:
This fixes part of #65

tests/packages/media-sound/bitwig-studio-1.3.16.conf
mrVanDalo: sure but than I have to use line breaks and it looks shit.

simonvanderveldt: But the above line is just 69 characters? No need for linebreaks.

mrVanDalo: Actually I don't like the echo "..." > /adsfasd syntax here.
If I want to add something to the file, I have to rewrite it.

simonvanderveldt: > If I want to add something to the file, I have to rewrite it.

No, in that case you can just use `>>` instead of `>`

mrVanDalo: > No, in that case you can just use `>>` instead of `>`

I consider this bad coding style.

mrVanDalo: I must ask, is THIS really a blocker ? 
I mean we are talking about code style here, on a level that is quite detailed.

mrVanDalo: @simonvanderveldt ping

simonvanderveldt: @mrVanDalo Sorry, was on holidays :)

> I must ask, is THIS really a blocker ?
> I mean we are talking about code style here, on a level that is quite detailed.

IMHO it is, because coding style only works if it's used consistently.
The shell script style guide says to use heredocs in case the line longer than 80 characters. It isn't in this case, so we should just use the simpler form.

mrVanDalo: > coding style only works if it's used consistently.
I would disagree. coding style is still a thing left to the developer. Following blindly rules which pop up to quickly, will not result in good code quality.

> The shell script style guide says to use heredocs in case the line longer than 80 characters. It > isn't in this case, so we should just use the simpler form.
As you point out the docs talk about lines which are longer than 80 chars. But it says nothing about lines which are shorter.

I still don't see a point to use `echo` here. It will not improve readability.

simonvanderveldt: Quote from the styleguide, emphasis mine
> **If** you have to write strings that are longer than 80 characters, this should be done with a here document or an embedded newline if possible.

So using heredocs is only suggested in case you have more than 80 characters on a line. In other words, not for cases with less than 80 characters on a line.

mrVanDalo: > In other words, not for cases with less than 80 characters on a line.

You can't reverse the sentence for the opposite situation.
"When it's raining it's getting cold" This does not imply "When it's not raining it's not getting cold"

mrVanDalo: @marcan It would be nice if you could also join the discussion, so we can achieve some kind of democratic consents here.

Or we could make a quick discussion about style-guides and such in the irc.

simonvanderveldt:
For future reference bitwig calls `jack_get_version_string` which is listed as a symbol that's only available in jack2, see https://github.com/jackaudio/jackaudio.github.com/wiki/JackAPIchanges12#symbols-present-in-jack2-but-missing-in-jack1

simonvanderveldt:
I'm still of the opinion the changes don't adhere to the styleguide so tso they should be fixed, but we need to move on, the tests are failing.

PR #83, opened 2017-09-05
Bump v1 suite to 0.8.4

simonvanderveldt:


simonvanderveldt:
> Why did you delete the old version ? maybe there this release it not as stable as the old version.

We did the same last time, didn't expect we'd want to keep the old version.
I can also just add the new one and keep the existing one. Would you prefer we do that?

simonvanderveldt:
@mrVanDalo I've updated the commits so they now add the 0.8.4 version instead of replacing the 0.8.3 version.

PR #87, opened 2017-09-10
media-plugins/distrho-ports: fix repoman QA warning for use of spaces instead of tabs

simonvanderveldt:
No actual changes to the ebuild, just changed spaces to tabs


simonvanderveldt:
> lol indention wars.

:P
Tabs are actually required for ebuilds, that's why repoman complains about them. See https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#indenting-and-whitespace

PR #88, opened 2017-09-11
Add virtual/liblash in order to satisfy portage

domichel:
- lash and ladish cannot be installed at the same time due to file collisions
  and portage want media-sound/lash.

This commit provide a lash-original ebuild and a fake lash ebuild. The fake lash ebuild install nothing but satisfy portage, which packages hard-depend on media-sound/lash and will block ladish.

The lash-original ebuild install the last lash version, and as its versioning is greater than portage's lash, it will be no maintenance of it until hopefully portage introduce ladish.

Our packages need to depend on virtual/liblash.


domichel:
Because you don't care if an user want to install a portage package which depend on lash. He will not only get a blocker between lash and ladish, but he will also not be able to install ladish at the same time than the wanted portage package.

domichel:
I made the ABI checker test and put the result on the bug. It look like they are not compatible. Which imply a virtual is not the right solution according to the bug. https://bugs.gentoo.org/show_bug.cgi?id=442916

simonvanderveldt:
This is the same things as answered in #191 

Quoting here for peopling finding this issue through the search:
> If you want to use packages that have LASH support and you have LADISH installed the easiest way to achieve that is to add `media-sound/lash` to `/etc/portage/package.provided`

PR #91, opened 2017-09-17
New and updated ebuilds

nick87720z:
Already made enough commits, before could get answer about pull request policy.
After all, it is never late to recreate.

### new
media-sound/carla
media-sound/eko
media-sound/jack-stdio - for specmatch dependency (need to test build flow before push)

### update
media-sound/stretchplayer - fixed, was removed due to build failure
media-sound/ladish - problem with dependencies
media-sound/a2jmidid
media-sound/xjadeo-0.8.7 - fixed ebuild, more use flags
media-sound/guitarix2 - update & version bump

### drobilla stuff (unbundled, use https, more use flags)
ganv, patchage, suil, serd, sord, sratom, lilv, raul
media-sound/lv2 - use https for git


dev-libs/serd/metadata.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer>
+		<email>proaudio@tuxfamily.org</email>
+		<name>pro-audio overlay team</name>
```
mrVanDalo: your name here please.

dev-libs/sord/sord-9999-r1.ebuild
```diff
@@ -0,0 +1,46 @@
...
+    doc? ( app-doc/doxygen )"
+
+DOCS=( AUTHORS NEWS README )
+
+src_configure() {
+	waf-utils_src_configure \
+		--docdir=/usr/share/doc/${PF} \
+		$(usex test   '--test --verbose-tests' '') \
+		$(usex doc         --docs '') \
+        $(usex static-libs --static '') \
```
mrVanDalo: looks like wrong indention.

simonvanderveldt:
@nick87720z I'll do a quick scan of all the ebuilds and give some high-level comments after that, OK?

simonvanderveldt:
After some trying an not really knowing what to do with so many changes, here's the best I can do:
- First of all, thanks for the PR :)
- Second point: this is too much stuff at once, which makes it difficult to give decent feedback
- Packages that are already in the gentoo tree should be fixed/updated/etc there. You can make a PR here to do so https://github.com/gentoo/gentoo (or file a bug at bugs.gentoo.org first). It seems like this covers the majority of packages/change of this PR
- ladish already exists in the overlay
- The maintainer should be the person maintaining the package, i.e. someone part of the [team](https://github.com/orgs/gentoo-audio/people). I don't think we want to do things like proxy-maintaining, but maybe someone else has a different opinion.
- Something on your end is wrong because your `Manifest` files have the files that live in the repo in them, this should not be the case for an overlay that has thin-manifests enabled (like this one has)
- Same for changelog files, they shouldn't exist in a git based overlay

I'll close this one for now, but feel free to ask questions if you want/need more info.

PR #93, opened 2017-10-01
x11-libs/ntk add datestamped version

simonvanderveldt:
Since we want our package to be in testing `~` we need the deps to be in testing as well, forgot about that when I added ntk :P
So this adds a datestamped version for ntk keyworded for testing :)

There were two changes this year, last one on 21.04.2017, no open bugs/issues since then so using that commit.


simonvanderveldt:
Closing because an NTK release has been tagged :)
I'll create a new PR to add that released version.

PR #103, opened 2017-10-21
media-libs/libgig: added live ebuild

fordfrog:
created live ebuild for libgig as linuxsampler does not work with version 4.0.0 which is in the main tree


fordfrog:
i have also attempted to get this ebuild into the main tree:
https://bugs.gentoo.org/show_bug.cgi?id=635050

mjeveritt:
Merged in https://github.com/gentoo/gentoo/commit/7f18041201c07add5cfff5007a64ba632dc10912 , https://bugs.gentoo.org/635050#c1 . Close this issue?

fordfrog:
@mjeveritt right, thanks!

PR #104, opened 2017-10-21
media-sound/linuxsampler: added live ebuild

fordfrog:
linuxsampler live ebuild

### DIFF
--- a/media-sound/linuxsampler/linuxsampler-9999.ebuild
+++ b/media-sound/linuxsampler/linuxsampler-9999.ebuild
@@ -1,0 +1,47 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+inherit eutils subversion
+
+DESCRIPTION="LinuxSampler is a software audio sampler engine with professional grade features"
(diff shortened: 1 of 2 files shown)


fordfrog:
yes, it will fail until libgig-9999 is available

fordfrog:
i have also attempted to get this ebuild to the main tree:
https://bugs.gentoo.org/show_bug.cgi?id=635052

simonvanderveldt:
Thanks for the PR! Is it OK if we wait until there's a response in the upstream bug?

fordfrog:
sure. for myself i have it in my dev overlay, so here it's for the others if they need it (or they can add my overlay for the meantime). it would make no sense to put it here if it would be put in the main tree, on the other side, it might take some time before it reaches the main tree. anyway, people have means to use it if they want.
and the same applies for libgig.

simonvanderveldt:
So, no response from upstream so far :(
@fordfrog What would you like to do? I noticed there's been a release of linuxsampler in the meantime http://doc.linuxsampler.org/Release_Notes/LinuxSampler_2_1_0/, maybe that bump would be enough?

fordfrog:
@simonvanderveldt  well, i'm using the live ebuild myself from my repo, what you want to add to your repo is your decision :-)

PR #110, opened 2017-10-28
add media-libs/libmonome

simonvanderveldt:
libmonome is needed to work with monome grid's https://monome.org/grid/
Apparently no one so far made an ebuild for it so here we go :)

Once merged I'll create a PR to update the monome Linux docs so people know they can install libmonome easily when using gentoo :)


simonvanderveldt:
Validated working using serialosc for which I'll create a PR soon

PR #138, opened 2017-12-09
Add media-libs/zita-resampler

simonvanderveldt:
Needed for among other things guitarix.

~I expect the emerge test to fail because the current stage3 is broken (no default profile defined).~
[edit] This has been fixed upstream, so it should be working again :)


fordfrog:
you can find working guitarix2 in my overlay at https://cgit.gentoo.org/dev/fordfrog.git/

simonvanderveldt:
> you can find working guitarix2 in my overlay at https://cgit.gentoo.org/dev/fordfrog.git/

@fordfrog thx for the link!
I ran into some issue with the wscript, which automatically does things even if you ask it not to (use the built in convolver/resampler for example when you want it to use the system installed ones). I'll see if I can get those things fixed upstream, otherwise I'll probably create a new ebuild with some patches based on yours if that's ok.

fordfrog:
you can copy them and modify them as you need.

PR #167, opened 2018-02-11
Add qtopengl dependency

nick87720z:
Cadence supports opengl via PyQt4.QtOpenGL, for what there is use flag.


simonvanderveldt:
Thanks for the PR!

Since the Cadence components check if QtOpenGL exists and only use it if it does I think this should be an optional dependency. See for example: https://github.com/falkTX/Cadence/blob/dea02f6421f040f70377e13501ccee1b5d9e0658/src/catia.py#L39

Can you add an `opengl` USE flag and use that to add this dependency?

PR #175, opened 2018-03-18
Update cadence patch to apply cleanly

dequbed:


simonvanderveldt:
Hi @dequbed Thanks for the PR! Looking at the changes I assume emerging cadence currently fails because the patch no longer applies?

One small question, can you rebase this PR on current master? That should ensure the tests all pass.

If you have no experience with rebasing I can do it as well, but I can only do so if you [allow me to push to your PR](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

dequbed:
Correct, the ebuild is still valid and works, just the Makefiles changed
slightly and the patch didn't apply cleanly anymore.

I rebased on to the current master, thank you for you quick reply.

simonvanderveldt:
Thanks for the quick update, the patch works again, thanks for that as well :)

Looking at the Makefile we'll need to patch some other things as well because the compile settings are hardcoded as well :(

I did run into the following error when emerging it though, caused by the warning combined with `-Wall`
```
g++ -c ../widgets/pixmapdial.cpp -O3 -ffast-math -mtune=generic -msse -mfpmath=sse -Wall -Wextra -DNDEBUG -fPIC -std=c++0x -march=core-avx-i -O2 -pipe -DQT_NO_DEBUG -DQT_NO_DEBUG_STREAM -DQT_NO_DEBUG_OUTPUT -I../widgets -I/usr/include/qt5/QtWidgets -I/usr/include/qt5 -I/usr/include/qt5/QtGui -I/usr/include/qt5 -I/usr/include/qt5/QtCore -I/usr/include/qt5 -o ../widgets/pixmapdial.o
jackmeter.cpp: In function 'void reconnect_ports()':
jackmeter.cpp:137:32: warning: unused variable 'jRecPort1' [-Wunused-variable]
         if (jack_port_t* const jRecPort1 = jackbridge_port_by_name(jClient, "system:capture_1"))
                                ^
jackmeter.cpp:143:32: warning: unused variable 'jRecPort2' [-Wunused-variable]
         if (jack_port_t* const jRecPort2 = jackbridge_port_by_name(jClient, "system:capture_2"))
```
Are you seeing the same?

dequbed:
Yes, it appears Cadence is switching to use PyQt5.
I'll fix the ebuild and open a new PR as soon as I have.

simonvanderveldt:
> Yes, it appears Cadence is switching to use PyQt5.
> I'll fix the ebuild and open a new PR as soon as I have.

Yeah, noticed that as well.
I had some time to spare so I've already updated the ebuild :) See https://github.com/gentoo-audio/audio-overlay/pull/177

PR #176, opened 2018-03-18
Tests: disable posting repoman result to GitHub for 3rd party PRs

simonvanderveldt:
Otherwise the build fails because the secrets/credentials are missing


simonvanderveldt:
Also validated that the PR comment with repoman results still works for PRs from this repo :)

PR #177, opened 2018-03-19
media-sound/cadence switch to qt5

### LINKED ISSUE #184: media-sound/cadence has missing dependency on dev-python/PyQt5
Application will not build without dev-python/PyQt5. Please add to dependencies in ebuild.

simonvanderveldt:
Quiet some new deps because of the switch to Qt5, I hope Travis won't time out :x

Fixes #184


simonvanderveldt:
Aaand of course Travis timed out. I'll have a look if we can do some better caching or reduce the amount of dependencies

simonvanderveldt:
Validated locally that this works.

PR #189, opened 2018-05-02
cadence-9999: fix the python shebangs to the single python target

### LINKED ISSUE #187: cadence fail with No module named 'PyQt5'
On my system, python3.4 is the default, and PyQt5 is installed only for python3.5, which result into:

```
# cadence 
Traceback (most recent call last):
  File "/usr/share/cadence/src/cadence.py", line 25, in <module>
    import PyQt5
ImportError: No module named 'PyQt5'
```

To solve this, I changed all the "python3" tests and calls to "python3.5" into the launcher scripts in src/data.

domichel:
This fix #187

### DIFF
--- a/media-sound/cadence/cadence-9999-r4.ebuild
+++ b/media-sound/cadence/cadence-9999-r4.ebuild
@@ -15,9 +15,10 @@ SLOT="0"
 
 IUSE="-pulseaudio a2jmidid ladish"
 
-RDEPEND="virtual/jack
+RDEPEND="${PYTHON_DEPS}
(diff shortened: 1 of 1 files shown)


simonvanderveldt:
@domichel Thanks for the fix!
Since we're fixing this anyway, shouldn't we just get rid of the whole check if a specific Python version exists in the scripts like here https://github.com/falkTX/Cadence/blob/master/data/cadence#L30 ?
Otherwise one might still end up in the same situation you had because another Python version can be used.

PR #191, opened 2018-05-18
media-sound/ladish: make compatible with lash (wrapper)

nick87720z:
There is package media-sound/lash-9998 from proaudio, intended to be wrapper for packages, supporting lash.
While proaudio version doesn't have this restricting dependency, i had it in local overlay in proposed form, and it is ok.

### DIFF
--- a/media-sound/ladish/ladish-9999.ebuild
+++ b/media-sound/ladish/ladish-9999.ebuild
@@ -31,7 +31,7 @@ RDEPEND="media-libs/alsa-lib
 	media-sound/jack2[dbus]
 	sys-apps/dbus
 	dev-libs/expat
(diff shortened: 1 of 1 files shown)


simonvanderveldt:
@nick87720z After the discussion on IRC: OK if I close this one? LASH and LADISH cannot be installed at the same time since they install the same files.

If you want to use packages that have LASH support and you have LADISH installed the easiest way to achieve that is to add `media-sound/lash` to `/etc/portage/package.provided`. When LADISH gets [added to the main tree](https://bugs.gentoo.org/442916) this obviously won't be necessary.

nick87720z:
ok

PR #196, opened 2018-06-29
media-sound/cadence replace virtual/jack dependency with media-sound/jack2

### LINKED ISSUE #195: media-sound/cadence needs two dependencies added to ebuild
These may seem trivial to most, but for those of us using CloverOS (or other minimal gentoo distributions), dbus is not built-in by default to most packages. Therefore, the bare minimum cadence requires is media-sound/jack2 built with dbus, and dev-python/PyQt5 built with dbus, in order to operate with more than greyed-out buttons. Accordingly, these need to be added as dependencies in the ebuild.

simonvanderveldt:
Because cadence relies on DBus and DBus support isn't currently available in JACK1
Also add missing dbus USE flags for PyQT5 and a2jmidi, both are required for Cadence to work

Fixes #195

### DIFF
--- a/media-sound/cadence/cadence-9999-r4.ebuild
+++ b/media-sound/cadence/cadence-9999-r4.ebuild
@@ -15,10 +15,10 @@ SLOT="0"
 
 IUSE="-pulseaudio a2jmidid ladish"
 
-RDEPEND="virtual/jack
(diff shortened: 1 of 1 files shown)


simonvanderveldt:
@jason-oliveira does this look good to you?

simonvanderveldt:
Validated locally that emerging works fine. Unfortunately the build takes longer than 50 minutes so Travis times out.

PR #197, opened 2018-06-30
tests: allow package specific test config without version

simonvanderveldt:
Both `./packages/<category>/<PN>.conf` and `./packages/<category>/<P>.conf` are sourced now.

This simplifies the test configs quiet a bit, since the majority of the package specific configs work for all versions of a package.
In case a version specific config is necessary that's still possible.


simonvanderveldt:
repoman isn't happy with the Qt4 deps of the vee one suite, not directly related to this PR
I'll fix that next :)

PR #199, opened 2018-06-30
media-sound/cadence add opengl USE flag

simonvanderveldt:
And make it depend on PyQt5's OpenGL support

Fixes #167


simonvanderveldt:
Unfortunately another timeout on Travis, so validated locally again, emerging works fine.
Also validated that with the `opengl` USE flag the OpenGL settings are no longer greyed out, so works as it should :)

PR #217, opened 2018-08-19
media-sound/cadence fix Python dependencies

### LINKED ISSUE #187: cadence fail with No module named 'PyQt5'

simonvanderveldt:
Also replace python version used in executable with scripts with

Fixes #187
Fixes #189 

This will probably time out on CI, it has to build some big Qt packages
[edit] And it did indeed time out. Validated locally, emerge is working fine, no QA issues.

### DIFF
--- a/media-sound/cadence/cadence-9999-r6.ebuild
+++ b/media-sound/cadence/cadence-9999-r6.ebuild
@@ -15,15 +15,31 @@ SLOT="0"
 
 IUSE="-pulseaudio a2jmidid ladish opengl"
 
-RDEPEND="media-sound/jack2[dbus]
+RDEPEND="${PYTHON_DEPS}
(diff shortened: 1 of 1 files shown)


simonvanderveldt:
> Not a huge fan of how upstream does things here, but this looks fine.

Yeah, it's definitely less than ideal. I tried to fix some things in the past, some of that got merged other issues didn't really get much understanding unfortunately.

PR #219, opened 2018-09-09
Add media-sound/carla

simonvanderveldt:
Adds http://kxstudio.linuxaudio.org/Applications:Carla

This is based on the version I've had in my personal overlay for some time (https://github.com/simonvanderveldt/simonvanderveldt-overlay/tree/master/media-sound/carla), but updated based on the rather large amount of changes that have happened during all the RCs for 2.0

Since there's no release yet of 2.0 this currently only contains a live/9999 ebuild, will add a versioned one once 2.0 is released.


simonvanderveldt:
> Looks good to me! Sorry for taking forever to review this.

Np, thanks for having a look at it! :)

PR #233, opened 2018-10-21
media-sound/cadence use correct homepage URL

simonvanderveldt:
Noticed the homepage was set incorrectly to the main KXStudio page instead of the cadence page :)


simonvanderveldt:
Travis timed out again :(

PR #288, opened 2019-11-16
tests: Only run repoman tests for new or changed ebuilds when testing a PR

simonvanderveldt:


simonvanderveldt:
The repoman failures are unrelated to this change (they were the reason for this change actually :P), I've validated these changes as working correctly locally.

PR #290, opened 2019-11-19
add media-plugins/adlplug

spinnylights:
Okay, here we go! Let me know if any changes are needed. It turned out to be more of a bear to write then I was hoping :stuck_out_tongue: but everything seems fine now. It passes the docker tests and I've been happily using the plugins I built from it locally. :smile_cat:

### DIFF
--- a/media-plugins/adlplug/adlplug-9999.ebuild
+++ b/media-plugins/adlplug/adlplug-9999.ebuild
@@ -1,0 +1,123 @@
+# Copyright 2019 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake-utils multibuild git-r3
+
+DESCRIPTION="FM synthesizer plugins based on OPL3 and OPN2 sound chip emulations"
+HOMEPAGE="https://github.com/jpcima/ADLplug"
+
+EGIT_REPO_URI="https://github.com/jpcima/${PN}.git"
+if [[ ${PV} == *9999 ]]; then
+	KEYWORDS=""
+else
+	# Upstream release tarballs do not contain the necessary third party libs;
+	# these are available via git submodules per upstream build instructions
+	EGIT_COMMIT="v${PV}"
+	KEYWORDS="~amd64"
+fi
+
+LICENSE="Boost-1.0 BSD GPL-3 GPL-2+ LGPL-3 LGPL-2.1 MIT ISC"
+SLOT="0"
+
+# Upstream offers a VST3 build option, but unfortunately JUCE 5.4.1
+# doesn't support VST3 on Linux
+# see https://github.com/WeAreROLI/JUCE/issues/307
+IUSE="+opl3 +opn2 +lv2 +vst jack alsa pch assertions"
+REQUIRED_USE="
+	|| ( opl3 opn2 )
+	|| ( lv2 vst jack alsa )
+"
+
+# JUCE will stop the build if alsa-lib is not present, regardless
+# of the build options
+DEPEND="
+	x11-libs/libXinerama
+	x11-libs/libXext
+	x11-libs/libXrandr
+	x11-libs/libXcursor
+	media-libs/freetype
+	media-libs/alsa-lib
+	jack? ( virtual/jack )
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	dev-util/cmake
+	|| (
+		>=sys-devel/gcc-4.8.1
+		>=sys-devel/clang-3.3
+	)
+"
+
+src_prepare() {
+	MULTIBUILD_VARIANTS=()
+	use opl3 && MULTIBUILD_VARIANTS+=( opl3 )
+	use opn2 && MULTIBUILD_VARIANTS+=( opn2 )
+
+	CMAKE_USE_DIR=${WORKDIR}/${P}
+	BUILD_DIR=${CMAKE_USE_DIR}/build
+
+	# see https://github.com/jpcima/ensemble-chorus/issues/10
+	cd ${CMAKE_USE_DIR}/thirdparty/JUCE
+	eapply "${FILESDIR}/gcc9_juce_fix.patch"
+	cd ${CMAKE_USE_DIR}
+
+	# Upstream build overwrites ALSA standalone binary with JACK
+	# standalone binary during install if both are enabled
+	# This patch installs the ALSA standalones as [name]-alsa
+	if use jack && use alsa ; then
+		eapply "${FILESDIR}/jack_and_alsa_standalones.patch"
+	fi
+
+	cmake-utils_src_prepare
+}
+
+adl_configure() {
+	CMAKE_BUILD_TYPE="Release"
+	local adl_prefix="ADLplug_"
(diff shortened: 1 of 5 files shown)


spinnylights:
Sorry for all the extra pseudo-commits showing in this PR. I'm not really sure what's causing them; I had to unshallow my local copy of the repo in order to push my changes so it might have something to do with that. I tried the solution [here](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch) but it doesn't seem to have helped. Let me know if you need me to recreate the PR.

simonvanderveldt:
Thx for the PR.
Can you rebase it on current master? Alternativele you can give me push access to your PR's branch then I can do it, or I can pull in only the relevant changes from your commits into a new branch as well.

spinnylights:
I think that should do the trick? I went ahead and gave you push access anyway just in case, though.

simonvanderveldt:
Thanks! And sorry for the delay, pretty much only have some spare time in the weekends.

The history still seems broken. Before I make some changes (which will mean a force push to your branch, assuming that works) I was wondering if you are using the gentoo mirrored version of this repo as the source maybe? Looks like it from the commits that are in there.

I'l post some inline questions as well.

media-plugins/adlplug/adlplug-9999.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	MULTIBUILD_VARIANTS=()
+	use opl3 && MULTIBUILD_VARIANTS+=( opl3 )
+	use opn2 && MULTIBUILD_VARIANTS+=( opn2 )
+
+	CMAKE_USE_DIR=${WORKDIR}/${P}
+	BUILD_DIR=${CMAKE_USE_DIR}/build
+
+	# see https://github.com/jpcima/ensemble-chorus/issues/10
+	cd ${CMAKE_USE_DIR}/thirdparty/JUCE
+	dos2unix "modules/juce_graphics/colour/juce_PixelFormats.h"
```
simonvanderveldt: It seems like the `dos2unix` invocation here is only required to make the patch work. Is that correct? If so I think it's cleaner to fix the patch directly instead of running `dos2unix` first.

media-plugins/adlplug/adlplug-9999.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	CMAKE_BUILD_TYPE="Release"
+	local adl_prefix="ADLplug_"
+	local mycmakeargs=(
+		"-D${adl_prefix}CHIP=${MULTIBUILD_VARIANT^^}"
+		"-D${adl_prefix}LV2=$(usex lv2)"
+		"-D${adl_prefix}VST2=$(usex vst)"
+		"-D${adl_prefix}Standalone=$(usex alsa)"
+		"-D${adl_prefix}Jack=$(usex jack)"
+		"-D${adl_prefix}PCH=$(usex pch)"
+		"-D${adl_prefix}ASSERTIONS=$(usex assertions)"
```
simonvanderveldt: Would there ever be a reason to want these two options for a normal user? Can't we simply leave both of them out/disabled?

media-plugins/adlplug/adlplug-9999.ebuild
```diff
@@ -0,0 +1,126 @@
...
+
+LICENSE="Boost-1.0 BSD GPL-3 GPL-2+ LGPL-3 LGPL-2.1 MIT ISC"
+SLOT="0"
+
+# Upstream offers a VST3 build option, but unfortunately JUCE 5.4.1
+# doesn't support VST3 on Linux
+# see https://github.com/WeAreROLI/JUCE/issues/307
+IUSE="+opl3 +opn2 +lv2 +vst jack alsa pch assertions"
+REQUIRED_USE="
+	|| ( opl3 opn2 )
```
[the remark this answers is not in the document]
spinnylights: The reason I wrote the ebuild this way is because of how the upstream build process works. There's no way to build both plugins in one go; you have to run cmake twice, changing the value of a parameter for each one (see [here](https://github.com/jpcima/ADLplug/blame/master/README.md#L73)).

As long as the build process has to run twice anyway, I figured we might as well give users the option to build only one of them if someone wants to. I think you're right that most users would always build both, which is why I made that the default.

I'd much rather have them in one package then split into two. Arch used to have them in two separate packages and I remember I found it cumbersome. It looks like they've made them into a split package now (which basically does the same thing this ebuild does; they share a single PKGBUILD).

media-plugins/adlplug/adlplug-9999.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	x11-libs/libXcursor
+	media-libs/freetype
+	media-libs/alsa-lib
+	jack? ( virtual/jack )
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	dev-util/cmake
+	|| (
+		>=sys-devel/gcc-4.8.1
```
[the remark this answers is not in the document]
spinnylights: The only reason I listed them is because the build requires a compiler with C++11 support (see [here](https://github.com/jpcima/ADLplug/blame/master/README.md#L60)). Those are the minimum versions of GCC and Clang that support it; those lines just specify that one or the other needs to be there. Since they're quite old now, I imagine most users have more recent versions; I just figured it wouldn't hurt to specify since Gentoo is so flexible. If you think it's unnecessary I'm not opposed to taking that part out.

Conversely, I left out icc because when I wrote the ebuild I thought it didn't have full C++11 support. More recently I found out that it does now, at least on Linux. So, we could theoretically support it too here if you think it's worth leaving this part in (although I don't have a copy to test the build with).

media-plugins/adlplug/adlplug-9999.ebuild
spinnylights: When I wrote this, I was concerned about git's autocrlf behavior changing the line endings of the patch during fetching/unpacking, but I realized that was only an issue because of my personal gitconfig. I've fixed the patch and taken out the `dos2unix` stuff.

media-plugins/adlplug/adlplug-9999.ebuild
spinnylights: Probably not—I just figured there was no harm in making them available, since upstream supplies them. They're off by default. There are some other packages that make a `pch` USE flag available; I don't really see the point outside of a development context but I didn't want to step on anyone's toes if they have a need for it. Similarly, I don't know why anyone would want assertions if they're not doing a debug build but I didn't want to second-guess anyone. We can take them out if you like.

spinnylights:
Goodness, sorry it took me so long to get around to this! I've been pretty swamped too. I think my repo should be properly in sync with yours now—you were right, I was working from the mirror on accident. I've done my best to respond to your comments; let me know if there are further changes you'd like me to make.

simonvanderveldt:
@spinnylights It's been way to long, sorry about that, but if you're still interested in this I've made an updated PR for this https://github.com/gentoo-audio/audio-overlay/pull/439
I kept your commit, so the attribution is still there, just made some changes to it. Also split the package in two, seemed like the simplest way to go.

PR #300, opened 2019-12-23
media-sound/bitwig-studio: Bump to 3.1

klardotsh:
https://www.bitwig.com/en/19/bitwig-studio-3_1.html


simonvanderveldt:
Thanks for the PR! :+1: 
Just checked it locally and it seems to be working as expected, no real changes vs 3.0.3 :)

PR #305, opened 2020-01-01
media-sound/cadence: Remove skip stripping patch

aqxa1:
Functionality is now included in upstream. See: https://github.com/falkTX/Cadence/commit/233e8c10cf65dbe51c25a8878c4c0fdf1ae39e7d


simonvanderveldt:
Ah, yeah, the [PR](https://github.com/falkTX/Cadence/pull/159) was merged, we also had an emerge test failure because of that today. Thanks for the quick fix!

PR #317, opened 2020-02-13
media-sound/mididings: Fix libboost_python<xy> detection

### LINKED ISSUE #312: media-sound/mididings-99999999: cannot find -lboost_python
Issue is that there is no exactly -lboost_python.
If it were -lboost_python27, -lboost_python36 etc, it would probably work.

simonvanderveldt:
Fixes #312 

@nick87720z Could you check if this works for you as well?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,17 @@ jobs:
           key: portage-pkgdir-repoman-{{ epoch }}
           paths:
             - ~/.portage-pkgdir
+  repoman new or changed ebuilds:
+    machine: true
+    steps:
+      - checkout
+      - restore_cache:
+          key: portage-pkgdir-repoman
+      - run: ./tests/repoman-new-or-changed-ebuilds.sh
+      - save_cache:
+          key: portage-pkgdir-repoman-{{ epoch }}
+          paths:
+            - ~/.portage-pkgdir
   emerge new or changed ebuilds:
     machine: true
     steps:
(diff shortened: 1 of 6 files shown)


nick87720z:
I see patch is applied, but it still crash on first attempt to link with -lboost_python (according to log).
I would suspect unnoticed src/SConstruct file, but can't find any clues to scons in distutils part:
```
setup.py:151:boost_python_suffixes = ['-%d.%d' % sys.version_info[:2]]
setup.py:153:    boost_python_suffixes.append('3')
setup.py:154:libraries.append(boost_lib_name('boost_python', boost_python_suffixes))
src/SConstruct:44:        boost_lib_name('boost_python'),
src/SConstruct:45:#        boost_lib_name('boost_python-py32'),
```

simonvanderveldt:
@nick87720z Hmm, that's strange. Can you post the output of the emerge as well as the output of `emerge --info`?

nick87720z:
Please try this, it seems to work (probably gentoo changed boost_python naming in boost):
```
diff --git a/setup.py b/setup.py
index 6376278..ba73565 100755
--- a/setup.py
+++ b/setup.py
@@ -148,7 +148,7 @@ sources = [
 
 include_dirs.append('src')
 
-boost_python_suffixes = ['-py%d%d' % sys.version_info[:2]]
+boost_python_suffixes = ['%d%d' % sys.version_info[:2]]
 if sys.version_info[0] == 3:
     boost_python_suffixes.append('3')
 libraries.append(boost_lib_name('boost_python', boost_python_suffixes))
```

nick87720z:
@simonvanderveldt I was just after you :). Already tested in separate clone. Your change results to something like -lboost_python-36, which is of course not found.

simonvanderveldt:
> @simonvanderveldt I was just after you :). Already tested in separate clone. Your change results to something like -lboost_python-36, which is of course not found.

That's very strange, because the hyphen has been removed, see https://github.com/gentoo-audio/audio-overlay/pull/317/files#diff-e18fa9d4cca7db2d0c5a9de4db5f465cR19
I tested it both locally and on our CI and it works in both cases. Are you sure you're build from the branch from this PR?

nick87720z:
Probably because last fix is not merged yet. Never tried to play with branches in installed overlays :) .

nick87720z:
About copyright - I faced it when first tried to check specmatch.
Besides years update, there are now "Gentoo Authors".

simonvanderveldt:
@nick87720z OK, clear, thanks for checking.
The QA error was unrelated, still tweaking the CI setup a bit, should be fixed now :)

PR #322, opened 2020-02-18
Mazes 80

mazes-80:
A bunch of music related stuffs, from my local overlay.
This set is a draft as I didn't check dependencies.
Some ebuilds need updated versions of already existing packages, which in turn require fixes for other existing packages.
Locally all is consistent but this PR is not according this first set.

zrythm: as is meson will fail because of too old versions of dependencies


mazes-80:
errata: lilv and suil are already up to date in the gentoo tree, so zrythm should not complain

PR #346, opened 2020-06-08
Carla fixes

### LINKED ISSUE #328: 64-bit media-sound/carla is installing libraries in the wrong places.
Carla is installing the 64-bit libraries in /usr/lib/carla instead of /usr/lib64/carla, and likewise, the pkgconfig in /usr/lib/pkgconfig instead of /usr/lib64/pkgconfig.  While it's not stopping Carla from building and working correctly, it prevents the 64-bit LMMS from detecting Carla (if media-sound/carla is installed) in order to install the carla plugins.

aqxa1:
Carla hardcodes $(PREFIX)/lib, so correct this. Closes #328

Secondly, add some changes to respect EPREFIX. See [Prefix Project](https://wiki.gentoo.org/wiki/Project:Prefix). Not really a big deal, just some minor fixes to be in line with modern ebuilds.


simonvanderveldt:
Hi @aqxa1 Thanks for the PR and the fixes and sorry for the delayed response.

Could you rebase the PR to fix the merge conflict?
And could you apply the same changes to the 2.0.0 ebuild?
Also I think a revbump for the ebuilds is required for a change like this.

If it's too much trouble let me know, I can also do it.

aqxa1:
Okay, I've rebased the changes and also added 32-bit and Windows plugin bridge support taken from https://github.com/falkTX/Carla/issues/913#issuecomment-602289396. If that is too much just for this commit, I can rebase and add those to a separate PR (I'm pretty new to rebasing but I think I've got the hang of it).

simonvanderveldt:
> Okay, I've rebased the changes and also added 32-bit and Windows plugin bridge support taken from [falkTX/Carla#913 (comment)](https://github.com/falkTX/Carla/issues/913#issuecomment-602289396). If that is too much just for this commit, I can rebase and add those to a separate PR (I'm pretty new to rebasing but I think I've got the hang of it).

Looking at the list of commits and merge conflicts I'd guess the rebase didn't entirely work :P
It should just be `git pull upstream` (or whatever you've named this remote) followed by a `git rebase upstream/master` and fix the merge conflict. I'm not sure that will work now, from what GitHub shows it seems like you did a merge?

Regarding the 32-bit and Windows changes I'd prefer to do that in a separate PR. I won't be able to validate the Windows stuff as I don't have or use any of that and I don't understand the 32-bit option, but we can discuss that in the new PR.

aqxa1:
Haha, yeah I see the mess now. It's as you say, I think I was doing a traditional merge a while back and must have left it in a bad state.

I'm pretty tired right now so I'll sleep on it, and perhaps just start the PR from scratch if it's not fixable. And yeah, I'll stick to just the smaller changes for now, no issues there.

PR #362, opened 2020-08-11
Add zyn-fusion

marcan:
zyn-fusion uses a whole pile of vendored git submodules, and some mruby gems stuff. This is the best I could do to eliminate all the compile-time clones and have everything fetched properly in the ebuild.

Note that some of the hashes aren't even pinned by the root version of zyn-fusion (the gems stuff), so for those it's just "whatever was master right now".

I did my best to fix the install layout to fit LSB standards, and fixed the QA warnings.

The zynaddsubfx ebuild is just from upstream portage, with the USE flag and optional dep added. When built that way, it requires zyn-fusion to show its UI, but does not otherwise depend on the package at build time (it dynamically loads it).


simonvanderveldt:
Do we need to do something with this?
```

 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * /var/tmp/portage/media-sound/zynaddsubfx-3.0.5-r1/work/zynaddsubfx-3.0.5/rtosc/src/dispatch.c:293:16: warning: argument 2 null where non-null expected [-Wnonnull]

 * Please do not file a Gentoo bug and instead report the above QA
 * issues directly to the upstream developers of this software.
 * Homepage: http://zynaddsubfx.sourceforge.net/
```

marcan:
zyn-fusion is just the *GUI*. It's a dependency of zynaddsubfx[zest]; alone it does nothing. zynaddsubfx is still developed, as it's still the core of zyn-fusion. I don't know about the old GUI, but if that's deprecated/unmaintained then we could just drop those USE flags for zynaddsubfx and make zyn-fusion a hard dep? Not sure there's much benefit in that though, as long as upstream still has it and it works.

I think we can ignore that QA warning. The code is definitely broken there, but it's an upstream issue, inherited from the version in portage anyway, and I don't know if there's actually a way a user could hit that codepath. [https://github.com/fundamental/rtosc/blob/master/src/dispatch.c#L293](Upstream rtosc) still has the same code.

marcan:
Filed the QA warning [up-upstream](https://github.com/fundamental/rtosc/issues/46) anyway.

simonvanderveldt:
> zyn-fusion is just the _GUI_. It's a dependency of zynaddsubfx[zest]; alone it does nothing. zynaddsubfx is still developed, as it's still the core of zyn-fusion. I don't know about the old GUI, but if that's deprecated/unmaintained then we could just drop those USE flags for zynaddsubfx and make zyn-fusion a hard dep? Not sure there's much benefit in that though, as long as upstream still has it and it works.

We've checked this with fundamental (the Zyn dev) and the old UI is still supported and supposed to work. Only caveat is that for the plugin which UI to use is chosen at build time. To prevent confusion it's probably best to allow either using the old/ftlk UI or the new/fusion UI.
(you already know this, this is more for future reference for myself :P)

Arch has two packages, one called [zynaddsubfx](https://www.archlinux.org/packages/community/x86_64/zynaddsubfx/) and one called [zyn-fusion](https://aur.archlinux.org/packages/zyn-fusion/), where one blocks the other. I know this is technically not correct, but it does seem a bit simpler to me, especially for users.
On the other hand, because Arch is binary, they to make the choice for which UI to build upfront so they need two packages, for us that's a choice for our users, so maybe the setup as proposed in the PR (zynaddsubfx package with a USE flag for fusion that depends on the zyn-fusion UI) is the better choice.
@marcan What do you think?

media-sound/zyn-fusion/files/zyn-fusion-gcc10.patch
```diff
@@ -0,0 +1,11 @@
+--- mruby-zest-build-3.0.5/src/mruby-widget-lib/src/gem.c.old	2020-08-11 21:26:06.146444612 +0900
++++ mruby-zest-build-3.0.5/src/mruby-widget-lib/src/gem.c	2020-08-11 21:27:51.839394549 +0900
+@@ -290,7 +290,7 @@
+     return self;
+ }
+ 
+-const struct mrb_data_type mrb_nvg_context_type;
++extern const struct mrb_data_type mrb_nvg_context_type;
```
simonvanderveldt, line 8: Do you know if this has already been PRed upstream? Arch is carrying a similar patch https://aur.archlinux.org/cgit/aur.git/tree/gcc10.patch?h=zyn-fusion

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
```diff
@@ -0,0 +1,116 @@
...
+USE_RUBY="ruby25 ruby26 ruby27"
+PYTHON_COMPAT=( python{2_7,3_6,3_7,3_8,3_9} )
+
+inherit bash-completion-r1 ruby-single python-any-r1
+
+DESCRIPTION="Zyn-Fusion User Interface"
+HOMEPAGE="https://github.com/mruby-zest/mruby-zest-build"
+
+SUBMODULES=(
+	"4f57a1ef9f968e9d5eef53667c7960a2e98c9750 deps/mruby-complex https://github.com/pbosetti/mruby-complex"
```
simonvanderveldt, line 15: I'm wondering if it wouldn't be better to simply us git-r3 together with the [mruby-zest-duilb repo](https://github.com/mruby-zest/mruby-zest-build/blob/v3.0.2/.gitmodules)?
That seems a lot easier to maintain than this, in case of an update we'd have to update all these submodules which seems pretty annoying tbh.

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
```diff
@@ -0,0 +1,116 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+USE_RUBY="ruby25 ruby26 ruby27"
+PYTHON_COMPAT=( python{2_7,3_6,3_7,3_8,3_9} )
```
simonvanderveldt, line 7: Is `PYTHON_REQ_USE='threads(+)'` no longer required for waf?

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
```diff
@@ -0,0 +1,116 @@
...
+
+S="${WORKDIR}/mruby-zest-build-$PV"
+
+PATCHES=(
+"${FILESDIR}/zyn-fusion-gcc10.patch"
+"${FILESDIR}/zyn-fusion-qml-path.patch"
+)
+
+src_prepare() {
+	# Unbundle libuv: makefile and rake file
```
simonvanderveldt, line 67: I think it would be slightly clearer/better to do this as patch(es)? What do you think?

media-sound/zynaddsubfx/zynaddsubfx-3.0.5-r1.ebuild
```diff
@@ -0,0 +1,96 @@
...
+src_configure() {
+	append-cxxflags -std=c++11
+
+	local mycmakeargs=(
+		-DPluginLibDir=$(get_libdir)
+		$(cmake_use_find_package alsa Alsa)
+		$(cmake_use_find_package doc Doxygen)
+		$(cmake_use_find_package fltk FLTK)
+	)
+	if use zest ; then
```
simonvanderveldt, line 76: Shouldn't we have a exactly one of for `fltk` and `zest`?
Now you can have both set but you'll still always get `fltk` for the plugin UI, I think that's a bit confusing.

media-sound/zyn-fusion/files/zyn-fusion-gcc10.patch
marcan: Not sure; this is a common theme for gcc10 compat issues and I wouldn't be surprised if upstream hasn't heard of it yet. I can throw a PR over at them :)

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
marcan: That's not valid for non-live ebuilds; in particular, it's not just the top-level submodules, but also the mruby stuff starts cloning repos down the line if you don't feed them to it ahead of time, during src_compile, and that's a definite no-no. git-r3 fetches in src_unpack which is also no good for regular ebuilds. Non-live ebuilds need to have everything fetched in as a SRC_URI, they can't go out and hit the internet themselves.

The SUBMODULES array is formatted and sorted the same way as `git submodule status`, so at least it should be pretty easy to copy and paste a pile of new hashes for an update.

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
marcan: Huh, hadn't heard of this one. I guess it is. I'll add it.

media-sound/zynaddsubfx/zynaddsubfx-3.0.5-r1.ebuild
marcan: I was wondering that myself when I was writing the ebuild, but then I looked at the REQUIRED_USE documentation and it says it should be used sparingly:

https://devmanual.gentoo.org/general-concepts/use-flags/index.html#conflicting-use-flags

In particular it seems exclusions like that are more useful for dependencies, while for leaf level ebuilds it makes more sense to favour user intent (which, if they ask for zest, probably means they don't want fltk).

An einfo message to this effect might be in order, though.

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
marcan: This actually came from the ebuild I based mine off of. It's actually calling pkg-config here, so it wouldn't work as a static patch. However, it should be possible to defer the pkg-config call to be run inside the Makefile and Ruby file. I can try moving it to a patch.

marcan:
I think it doesn't make sense to embed zynaddsubfx into zyn-fusion itself. It would be duplicating work into two ebuilds, and complicate dependencies if e.g. an app somewhere decides it needs to depend on zynaddsubfx (the core plugin) for some functionality, regardless of GUI.

We could add a message to zyn-fusion that it is only the GUI, and that if users want to use it with zynaddsubfx they need to build zynaddsubfx[zest].

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
simonvanderveldt: Ah, you're right, I thought it just replaced the fixed values with `pkg-config` invocations but it replaces the values with the results of the `pkg-config` invocations, sorry about that.

media-sound/zyn-fusion/files/zyn-fusion-gcc10.patch
simonvanderveldt: OK. Would be nice to do that. Don't need to wait for it with this PR though :)

media-sound/zynaddsubfx/metadata.xml
```diff
@@ -0,0 +1,19 @@
...
+		<email>marcan@marcan.st</email>
+		<name>Hector Martin</name>
+	</maintainer>
+	<maintainer type="project">
+		<email>proaudio@gentoo.org</email>
+		<name>Gentoo ProAudio Project</name>
+	</maintainer>
+	<use>
+		<flag name="dssi">Enable support for DSSI Soft Synth Interface</flag>
+		<flag name="zest">Enable support for the zyn-fusion zest UI</flag>
```
simonvanderveldt, line 14: Shouldn't we call this USE flag fusion? Zest is more of an internal name I think?

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
simonvanderveldt: AFAIK git-r3 should work, it checks out submodules recursively. I've used it in the past as well without issues. I'll try it out.

simonvanderveldt:
> I think it doesn't make sense to embed zynaddsubfx into zyn-fusion itself. It would be duplicating work into two ebuilds, and complicate dependencies if e.g. an app somewhere decides it needs to depend on zynaddsubfx (the core plugin) for some functionality, regardless of GUI.

Yeah, agreed
 
> We could add a message to zyn-fusion that it is only the GUI, and that if users want to use it with zynaddsubfx they need to build zynaddsubfx[zest].

That would probably be a good idea. I'm not sure if there's a better solution (I guess we could let zyn-fusion depend on zynaddsubfx but that does mean that one would get zynaddsubfx as well even if you only want to run the UI locally, not sure how relevant that usecase is though. Also I guess that would mean a circular dependency with `zynaddsubfx[zest]` also depending on zyn-fusion).

media-sound/zynaddsubfx/metadata.xml
marcan: Sounds good, I'll change it

media-sound/zyn-fusion/zyn-fusion-3.0.5.ebuild
simonvanderveldt: Hmm, OK. That introduces a lot of extra work for maintaining packages, similar to the dependency versioning of go and rust packages. Gentoo/the gentoo devs are not being very practical here imho.
Anyway, I guess for now this means this is what it is? We'll just have to maintain a rather large list of dependencies and their versions/git commits?
What could/should we do to improve this? Ask upstream to create release tarballs that include all dependencies?

marcan: I dunno, I agree with the Gentoo devs that packages should be using proper distfiles-capable tarballs. What they do for packages that don't do this properly is roll their own tarballs. VCS ebuilds are fine for development, but they aren't nice for users... lots of subtle details break when you have an ebuild go out and pull things from the internet. For one, that loses the digest check, so you can't guarantee package authenticity (it being git gives you some of that back, but it's not perfect).

The correct solution is indeed for upstream to release proper tarballs with everything (not the github autoarchives). That's what `www-servers/h2o` has (another package using mruby), which is why it doesn't have this problem.

That said, it isn't really a lot of work. Just checkout the git tag, and paste the output of `git submodule status | cut -d" " -f2-3` in in block overwrite mode, if it's just commit changes for the existing set of deps.

marcan:
FYI, I have a busy week, so I'll be a bit late updating this, but I'm not forgetting :)

marcan:
I took care of the comments, but I'm at a loss as to the CMake issue since it doesn't happen here. Also, the zynaddsubfx ebuild is basically unchanged other than adding the dep and changing the UI mode, which seems unrelated to that Nio issue.

Can you check if you can build upstream zynaddsubfx from portage?

simonvanderveldt:
> I took care of the comments, but I'm at a loss as to the CMake issue since it doesn't happen here. Also, the zynaddsubfx ebuild is basically unchanged other than adding the dep and changing the UI mode, which seems unrelated to that Nio issue.
> 
> Can you check if you can build upstream zynaddsubfx from portage?

Yes, I'll give this a try this week and report back, will probably be Wednesday if that's OK.

simonvanderveldt:
> I took care of the comments, but I'm at a loss as to the CMake issue since it doesn't happen here. Also, the zynaddsubfx ebuild is basically unchanged other than adding the dep and changing the UI mode, which seems unrelated to that Nio issue.
> 
> Can you check if you can build upstream zynaddsubfx from portage?

I get the same error when building zyn from the gentoo tree. Which USE flags are you using?
```
 ❯ USE="jack" sudo -E emerge -1av zynaddsubfx                 

These are the packages that would be merged, in order:

Calculating dependencies... done!
[binary  N     ] dev-libs/mxml-3.1::gentoo  USE="-static-libs -threads" 1,391 KiB
[binary  N     ] x11-libs/libXpm-3.5.13::gentoo  USE="-static-libs" ABI_X86="(64) -32 (-x32)" 100 KiB
[ebuild  N     ] media-sound/zynaddsubfx-3.0.5::gentoo  USE="fltk jack -alsa -doc -dssi -lash -portaudio" 7,951 KiB
...
-- Will build NTK gui
CMake Error at src/Nio/CMakeLists.txt:28 (CHECK_INCLUDE_FILES):
  Unknown CMake command "CHECK_INCLUDE_FILES".
```

You might not run into this when JACK is not enabled, see https://github.com/zynaddsubfx/zynaddsubfx/blob/3.0.5/src/Nio/CMakeLists.txt#L23

simonvanderveldt:
FYI The upstream PR with this fixed has been merged into Zyn https://github.com/zynaddsubfx/zynaddsubfx/pull/82

mazes-80:
Notice to new upstream: if you have some problems with this ebuild you may ping me on irc. I sent those ebuilds to marcan a while ago, and he fixed some points that were in those.

SpomJ:
Can this at least be forwarded to GURU? Doesn't seem like it's getting accepted here anytime soon.

mazes-80:
Le Fri, 08 Nov 2024 03:59:41 -0800,
SpomJ ***@***.***> a écrit :

> Can this at least be forwarded to GURU? Doesn&#39;t seem like
> it&#39;s getting accepted here anytime soon.
> 

I think I can close this issue as last version of zynaddsubfx in portage
tree uses the new fusion interface.
I liked much the idea of being able to have both interfaces (gentoo is
about choice), but it was not really possible to have only fusion the
way those ebuild were thought, (only fltk, or both fltk+fusion was
supported)

PR #374, opened 2020-09-17
Add media-plugins/x42-plugins

simonvanderveldt:
Not sure if this should be in `media-sound` instead since it also installs some standalone JACK applications. `media-sound/x42-plugins` might be a bit of a confusing name though, so not really sure.


media-libs/libltc/libltc-9999.ebuild
```diff
@@ -0,0 +1,18 @@
...
+	inherit git-r3 autotools
+	EGIT_REPO_URI="https://github.com/x42/libltc.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://github.com/x42/libltc/releases/download/v${PV}/${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+LICENSE="LGPL-3"
+SLOT="0"
+RESTRICT="mirror"
```
marcan, line 18: Why this restriction? I thought this was only for licensing/bindist issues.

media-plugins/x42-plugins/x42-plugins-20200714.ebuild
```diff
@@ -0,0 +1,39 @@
...
+"
+DEPEND="${RDEPEND}
+	sys-apps/help2man"
+
+src_compile() {
+	emake STRIP="#" FONTFILE="/usr/share/fonts/dejavu/DejaVuSans-Bold.ttf"
+}
+
+src_install() {
+	emake DESTDIR="${D}" PREFIX="/usr" install
```
marcan, line 38: Probably `PREFIX="${EPREFIX}/usr"` is preferred here

media-plugins/x42-plugins/x42-plugins-20200714.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+DESCRIPTION="Collection of LV2 plugins"
+HOMEPAGE="https://github.com/x42/x42-plugins"
+SRC_URI="http://gareus.org/misc/x42-plugins/${P}.tar.xz"
+KEYWORDS="~amd64"
+LICENSE="GPL-2"
+SLOT="0"
+RESTRICT="mirror"
+
+RDEPEND="dev-libs/glib
```
marcan, line 14: I assume there are minimum version requirements for a lot of these deps... if upstream documents any of them they should be added here, if they don't I guess we don't have to care for now.

media-libs/libltc/libltc-9999.ebuild
simonvanderveldt: I think this is something that all overlays should be using because otherwise portage first tries to download a copy of the SRC_URI(s) from the gentoo mirrors, which obviously doesn't work for pretty much all our packages.
(This was already mentioned on IRC, figured I should post it here as well).

media-plugins/x42-plugins/x42-plugins-20200714.ebuild
simonvanderveldt: I'll do a check if I can find some specific versions.

media-plugins/x42-plugins/x42-plugins-20200714.ebuild
simonvanderveldt: Ah, yeah, I expect I'll forget the EPREFIX stuff many times :x I'll update the ebuilds to include this.

simonvanderveldt: Just looked into this and tbh I'd like to not do the prefix stuff.
Changes to ebuilds are tricky because of interactions between src_configure, src_compile and src_install as well as the EPREFIX/ED/EROOT variables. I have no reason to use it myself and we currently have no way to automatically test if it's correct. Also the packages in this overlay are pretty much targeted at a niche (Linux audio) within a niche (Gentoo). If someone wants Linux audio packages they should just install it in their system's OS.
We already barely have time to maintain what we have and to be frank I don't have the time to figure out stuff that at least to me is unclear and brings very limited value. Because of this I'd prefer to not do anything with regards to prefix.

marcan: My idea was to just dump `${EPREFIX}` in front of any paths passed to non-builtins and call it a day (and maybe use ED for any direct image accesses), since a bunch of upstream ebuilds do that too. If you feel uncomfortable doing that then it's fine not to do it; I just figured we might as well attempt to support such use cases even if we don't test it.

media-libs/libltc/libltc-9999.ebuild
marcan: Yeah, sounds reasonable :)

media-plugins/x42-plugins/x42-plugins-20200714.ebuild
simonvanderveldt: Checked this, there are only minimal versions defined for LV2, not for anything else. For LV2 it's version 1.8.1 which is long out of the gentoo tree, so there are effectively no version requirements.

marcan (approved):
Forgot to hit the approve button on this! Sorry!

PR #375, opened 2020-09-21
add media-plugins/dragonfly-reverb

marcan:


media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	sed -i -re '/^HAVE_(CAIRO|OPENGL|X11|JACK|LIBLO)/d' dpf/Makefile.base.mk || die 'sed failed'
+	# Do not strip
+	sed -i -re 's/-Wl,--strip-all//g' common/Makefile.mk
+}
+
+src_configure() {
+	echo "Nothing to configure"
+}
+
+src_compile() {
```
simonvanderveldt, line 41: Since this is using DPF, does this build all plugins formats (DSSI, LV2, VST)?
I added a bunch of build options for Nekobi (another DPF based plugin) to allow the choice of what to build, see here https://github.com/simonvanderveldt/Nekobi/commits/master
Not sure if we want this level of control/granularity or not? AFAIK DSSI requires liblo so if we can't control which plugin types are built we might need to add that as a dep (assuming it's a build-time dep).

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+DOCS=( README.md )
+
+src_prepare() {
+	default
+
+	# Remove automagic deps
+	sed -i -re '/^HAVE_(CAIRO|OPENGL|X11|JACK|LIBLO)/d' dpf/Makefile.base.mk || die 'sed failed'
+	# Do not strip
+	sed -i -re 's/-Wl,--strip-all//g' common/Makefile.mk
```
simonvanderveldt, line 34: Is this still necessary? I don't see any `strip` commands in this Makefile (https://github.com/michaelwillis/dragonfly-reverb/blob/3.2.1/common/Makefile)?

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+	cd "${S}/bin"
+
+	use jack && dobin $(ls -d Dragonfly* | grep -v '\.')
+
+	exeinto "/usr/$(get_libdir)/vst"
+	doexe Dragonfly*-vst.so
+
+	for i in Dragonfly*.lv2; do
+		exeinto "/usr/$(get_libdir)/lv2/${i}"
```
simonvanderveldt, line 66: Should this include `PREFIX`?

simonvanderveldt (commented):
Nice! Wanted to add this one myself as well :)
Added some questions/comments inline, nothing major.

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	EGIT_REPO_URI="https://github.com/michaelwillis/dragonfly-reverb.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://github.com/michaelwillis/dragonfly-reverb/releases/download/${PV}/DragonflyReverb-Source-v${PV}.tar.gz"
+	KEYWORDS="~amd64"
+	S="${WORKDIR}/DragonflyReverb-Source-v${PV}"
+fi
+LICENSE="GPL-3"
+SLOT="0"
+IUSE="jack"
```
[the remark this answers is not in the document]
marcan: SGTM!

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+RDEPEND="x11-libs/libX11
```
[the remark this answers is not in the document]
marcan: Nope. Just X/OpenGL.

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
marcan: This is an effect, so I guess you mean LADSPA, not DSSI. But as far as I can tell, this plugin always unconditionally builds LV2 and VST versions, and can optionally build standalone JACK versions.

I already get rid of HAVE_LIBLO, so even if you have it installed it won't try to use it. But even if I set it to true it won't build LADSPA versions anyway, so I think all of that is not used by DragonFly. I don't think there are any extra deps for the LV2/VST builds, so we might as well build them unconditionally.

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
marcan: I believe these functions already take into account `EPREFIX`. Nothing in portage prepends that when using them (rather, some ebuilds *strip out* EPREFIX from separately-computed paths before passing them).

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
marcan: It's not a command, it's a linker option. But you're right, this isn't necessary: this Makefile is only used for the `common` subdirectory, which only builds object files, no shared libs itself, so it never actually runs the linker. `SKIP_STRIPPING=true` below is necessary and sufficient to avoid stripping the final binaries.

media-plugins/dragonfly-reverb/dragonfly-reverb-9999.ebuild
simonvanderveldt: You're right, got confirmation from the #gentoo-prefix channel on IRC, it's in the docs.
from https://devmanual.gentoo.org/function-reference/install-functions/
> The following functions are provided by the package manager to handle various install-related tasks. ${ED} is prepended automatically to the install location. 

and from https://devmanual.gentoo.org/ebuild-writing/variables/index.html
> ED Shorthand for ${D%/}${EPREFIX}/

simonvanderveldt (approved):
LGTM! Also tested it locally and all working as expected :+1: 

Only minor thing would be adding RESTRICT=mirror because now I'm getting this before portage tries to download from the SRC_URI itself
```
>>> Emerging (1 of 1) media-plugins/dragonfly-reverb-3.2.1::audio-overlay
>>> Downloading 'http://ftp.snt.utwente.nl/pub/os/linux/gentoo/distfiles/2d/DragonflyReverb-Source-v3.2.1.tar.gz'
--2020-09-22 22:14:50--  http://ftp.snt.utwente.nl/pub/os/linux/gentoo/distfiles/2d/DragonflyReverb-Source-v3.2.1.tar.gz
Resolving ftp.snt.utwente.nl... 130.89.149.20
Connecting to ftp.snt.utwente.nl|130.89.149.20|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-09-22 22:14:50 ERROR 404: Not Found.

>>> Downloading 'http://mirror.leaseweb.com/gentoo/distfiles/2d/DragonflyReverb-Source-v3.2.1.tar.gz'
--2020-09-22 22:14:50--  http://mirror.leaseweb.com/gentoo/distfiles/2d/DragonflyReverb-Source-v3.2.1.tar.gz
Resolving mirror.leaseweb.com... 5.79.108.33
Connecting to mirror.leaseweb.com|5.79.108.33|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2020-09-22 22:14:50 ERROR 404: Not Found.
```

marcan:
Added RESTRICT=mirror before merging.

PR #380, opened 2020-09-27
media-sound/new-session-manager: Bump to 1.4.0

simonvanderveldt:
The flag to build the GUI has changed so I've updated that as well :)


marcan (approved):
Looks reasonable, assuming it builds/runs LGTM :+1:

simonvanderveldt:
> Looks reasonable, assuming it builds/runs LGTM +1

It does :)
image

I noticed 2 QA issues with ebuild (icon/.desktop file update was missing and the docs were installed in the wrong place), both are fixed now.

PR #382, opened 2020-09-30
media-plugins/distrho-ports: Use meson to build the live ebuild

### LINKED ISSUE #360: distrho-ports-9999 fails to build because it has switched to meson
The distrho-ports live ebuild fails because it has been switched to meson, need to update the ebuild.

simonvanderveldt:
Upstream has replaced make with meson

Fixes #360

### DIFF
--- a/media-plugins/distrho-ports/distrho-ports-99999999.ebuild
+++ b/media-plugins/distrho-ports/distrho-ports-99999999.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit git-r3
+inherit git-r3 meson
 
 DESCRIPTION="Linux ports of Distrho plugins."
 HOMEPAGE="https://github.com/DISTRHO/DISTRHO-Ports"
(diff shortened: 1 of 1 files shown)


media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -22,41 +22,26 @@ REQUIRED_USE="|| ( lv2 vst )"
...
-	if use !lv2; then
-		rm -rf "${D}"/usr/lib/lv2
-	fi
-	if use !vst; then
-		rm -rf "${D}"/usr/lib/vst
-	fi
-	rm -rf "${D}"/usr/src
+src_configure() {
+	local emesonargs=(
+		-Doptimizations=false
```
simonvanderveldt, line 41: This disables default for enabling SSE2/3 etc, see https://github.com/DISTRHO/DISTRHO-Ports/blob/master/meson.build#L129 I think this is what we should be doing, right? To allow the user to choose.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -22,41 +22,26 @@ REQUIRED_USE="|| ( lv2 vst )"
 
 RDEPEND="media-libs/alsa-lib
 	media-libs/freetype
-	media-libs/ladspa-sdk
```
simonvanderveldt, line 25: I don't really understand the situation with LADSPA but it seems like distrho-ports doesn't build LADSPA plugins anymore although there is mention of LADSPA stuff throughout the codebase. I'll ask Falk about this.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -22,41 +22,26 @@ REQUIRED_USE="|| ( lv2 vst )"
 
 RDEPEND="media-libs/alsa-lib
```
simonvanderveldt, line 23: There are some libs listed in the meson.build file that I don't know, any ideas how to determine what package they belong to?
https://github.com/DISTRHO/DISTRHO-Ports/blob/062a0f614703293bbc32eeda0fd65a297604262c/meson.build#L79

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -22,41 +22,26 @@ REQUIRED_USE="|| ( lv2 vst )"
...
-		rm -rf "${D}"/usr/lib/lv2
-	fi
-	if use !vst; then
-		rm -rf "${D}"/usr/lib/vst
-	fi
-	rm -rf "${D}"/usr/src
+src_configure() {
+	local emesonargs=(
+		-Doptimizations=false
+		$(meson_use vst build-vst2)
```
simonvanderveldt, line 42: I figured there wouldn't be much to gain from splitting this so I used a single `vst` USE flag for both vst2 and vst3

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
simonvanderveldt: Got an answer from Falk, there's no way to build LADSPA plugins, it's just references from the included Juce code.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
marcan: yeah, dl/rt are standard libc things (dynamic linker and realtime support)

marcan (approved):
Approving, modulo making sure that it's honoring user CFLAGS.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
simonvanderveldt: No, it's fully provided by the JUCE libs that are included

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
simonvanderveldt: Thanks! That was pretty simple, I should've simply checked the libs as you did, thanks for the examples!
Since these are provided by glibc I don't think we need any additional dependencies.

simonvanderveldt:
> modulo making sure that it's honoring user CFLAGS.

I'll have to check this, will report back once I've done so

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
simonvanderveldt: `CFLAGS` are honored, it's using the `-march=core-avx-i -O2` from my `make.conf` :)

PR #433, opened 2021-01-04
media-sound/setbfree: add missing GUI deps

marcan:
Package-Manager: Portage-3.0.12, Repoman-3.0.2


simonvanderveldt (approved):
LGTM!
I'm not sure if this requires a revbump.

marcan:
I think it does to make sure existing installed packages are rebuilt with the missing deps (otherwise what happened to me might happen, those deps getting depcleaned :)).

PR #439, opened 2021-01-25
Add media-plugins/adlplug and media-plugins/opnplug

### LINKED ISSUE #229: Add ADLMIDI and OPNMIDI plugins?
See https://github.com/jpcima/ADLplug

simonvanderveldt:
Fixes #229 
Closes #290


spinnylights:
Just wanted to note that I'm fine with this. If splitting it into two packages is the easiest way to go, I suppose we might as well do that, as long as it's easy for users to understand that they'll need to install both if they want both plugins. If it turns out that there's any other work that needs to be done on this, ping me and I'll happily chip in.

PR #446, opened 2021-02-06
media-sound/ladish: move to fork with python 3 compat

maxaudron:
I spent the weekend pulling ladish away from using waf to meson and
converting everything to python 3 as needed. The fork does not install
laditools I'd put that into a separate ebuild soon.

I'm having some smaller issues in conjunction with cadence right now
and it's not autostarting ladishd, don't know exactly why yet butt I think
it's more related to cadence.

And for the future I'm planning to bring back the GUI and potentially
 rewrite ladish_control in C to get rid of the python dep

### DIFF
--- a/media-sound/ladish/ladish-9999.ebuild
+++ b/media-sound/ladish/ladish-9999.ebuild
@@ -3,16 +3,16 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_6 python3_7 python3_8 )
+PYTHON_COMPAT=( python{3_6,3_7,3_8,3_9} )
 PYTHON_REQ_USE='threads(+)'
 
-inherit flag-o-matic python-single-r1 waf-utils
+inherit meson python-single-r1
 
 DESCRIPTION="LADI Session Handler - a session management system for JACK applications"
 HOMEPAGE="https://ladish.org"
 if [[ ${PV} == *9999 ]]; then
 	inherit git-r3
-	EGIT_REPO_URI="https://git.nedk.org/lad/ladish.git"
+	EGIT_REPO_URI="https://git.nedk.org/lad/${PN}.git"
 	KEYWORDS=""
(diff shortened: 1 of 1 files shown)


simonvanderveldt:
Thanks for the PR. Did you discuss this with nedko? He's pretty active at the #lad IRC. I pointed him at the openSUSE patches for Python3 (https://build.opensuse.org/package/show/multimedia:apps/ladish) just this week which he would look into.
I have most of those patches applied to the ebuild locally as well, but would prefer it to be fixed upstream and pull those commits in as patches.

maxaudron:
I talked to nedko and he is interest, but I'll have to clean up my stuff first, I'll update the build to the new upstream then.

simonvanderveldt:
FYI I've just merged some changes to make both ladish-1 as well as ladish-9999 work with Python 3, all of this still uses waf to build and the new upstream git repo from nedko. I didn't really want to switch to a third party repo, especially now that most of your changes have been merged into nedko's repo. He also said he expects to create a new release in March, if/when that happens we can look at switching to using meson, although I don't think he uses it himself, so it might not get the same level of attention as the waf build?

maxaudron:
Sorry, kinda fell off my table, I've had the ebuild lying around for a bit already. 

Gladish works fine with this too, and I've been using it for a bit, there are some parts that open to laditools which right now isn't working in py3 quite yet, but it's not much of a hard dependency, so I think it's ok.

simonvanderveldt:
> Gladish works fine with this too, and I've been using it for a bit, there are some parts that open to laditools which right now isn't working in py3 quite yet, but it's not much of a hard dependency, so I think it's ok.

Gladish is still GTK2 though, right? I haven't seen any commits that update it to GTK3.

maxaudron:
yes still gtk2

simonvanderveldt:
ok

PR #463, opened 2021-06-09
media-sound/bitwig-studio: Bump to 3.3.10

KevinThomas0:
Looks like there were just some small file name changes in `/usr/share`


simonvanderveldt:
Thanks for the PR!
There are two small QA issues, can you amend the PR/commit with fixes for them?

```
 * QA Notice: This package installs one or more .desktop files that do not
 * pass validation.
 * 
 * 	/usr/share/applications/com.bitwig.BitwigStudio.desktop: error: value "1.5" for key "Version" in group "Desktop Entry" is not a known version
 * 


 * QA Notice: Pre-stripped files found:
 * /opt/bitwig-studio/lib/cp/org/usb4java/linux-x86_64/libusb4java.so
 ```
I'm not sure about the version, it used to be 1.0, so we can always revert to that. Or just drop the whole line, not sure it serves any real purpose.
For the prestripped files maybe just try the whole `lib` directory or otherwise just add the whole `cp` directory.

KevinThomas0:
Thanks for the feedback!

> I'm not sure about the version, it used to be 1.0, so we can always revert to that. Or just drop the whole line, not sure it serves any real purpose.

Yeah I think we can just remove that, it seems not necessary: https://askubuntu.com/questions/827361/error-value-5-1-2-for-key-version-in-group-desktop-entry-is-not-a-known-v

simonvanderveldt:
Thanks for the update, looks good to me! :+1: 

Btw the version 1.5 is actually correct according to the freedesktop specifications, so I've created a bug for that https://bugs.gentoo.org/795570

PR #466, opened 2021-06-11
synthv1 drumkv1 padthv1 samplv1 now use cmake as default build system

mazes-80:
This PR now only contains synthv1


mazes-80:
nsm flag still untested

PR #469, opened 2021-06-14
media-plugins/nekobi: update patch

mazes-80:
Upstream updates Makefiles, so I updated patch.

Personnaly, I avoid patches in live ebuilds I rather try to add sed lines to src_prepare because they are more tolerant to file changes and simple offsetting than patches.

I know gentoo policy encourages patches for more readability, but for live ebuild this is non sense.
I don't think it improves readability for my part, but still stick to this policy for versionned ebuild.
Generally sed lines in live ebuilds helps me nicely to update versioned patches.


nick87720z:
Such sort of patches seems better be autogenerated with sed or awk.

PR #482, opened 2022-02-16
media-sound/{carla,cadence}: Fix PYTHON_USEDEP

NexAdn:
The ebuild used PYTHON_MULTI_USEDEP inside python_gen_cond_dep. However,
according to python-single-r1.eclass(5), only PYTHON_SINGLE_USEDEP and
PYTHON_USEDEP can be set, where PYTHON_USEDEP is probably what was
intended by the original ebuild maintainer.

This commit updates the ebuild by using PYTHON_USEDEP instead of
PYTHON_MULTI_USEDEP.

Package-Manager: Portage-3.0.30, Repoman-3.0.3


simonvanderveldt:
Thanks! :+1: 
I have the same changes locally, just haven't had the time to push them :x

PR #486, opened 2022-02-19
media-sound/bitwig-studio: Fix install

### LINKED ISSUE #480: Cannot install bitwig-studio
While installing `bitwig-studio`, the following error occurs:
~~~
chmod: cannot access '/var/tmp/portage/media-sound/bitwig-studio-4.0.4/image/opt/bitwig-studio/lib/jre/lib/*': No such file or directory
~~~

The `*` wildcard is not being expanded.

simonvanderveldt:
fperms seems to escape the wildcard argument when it passes it to chmod making the chmod fail
Work around this by simply no longer using doins which requires fixing the permissions but just cp-ing everything in opt/bitwig-studio
This the same thing that's being done for other binary packages in the gentoo tree like openjdk-bin and slack

Fixes #480


KevinThomas0:
Nice!

PR #499, opened 2022-03-05
add qpwgraph

xhub:
The author of qjackctl is working on a similar software for pipewire.
He started to advertise it, see https://www.reddit.com/r/linuxaudio/comments/t1w3b0/ann_qpwgraph_v021_released_patchbay_feature/
It is packaged by Arch.

AFAIK gentoo doesn't have Qt6 packaged. Also the Wayland "support" is disable. I still have X11 so I can't test it.

*Issue*: 
- The system tray does not feature an icon. 

Basic testing (changing connection graph) works fine.

Hope I nailed the PR process and I am happy to get feedback.


media-sound/qpwgraph/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>proaudio@gentoo.org</email>
+		<name>Gentoo ProAudio Project</name>
+	</maintainer>
```
NexAdn, line 7: We are not associated with the Gentoo ProAudio project and they don't maintain packages in this overlay. If you are interested in this package, please consider adding yourself as a `<maintainer type="person">` here as has been done by other people for other packages as well.

media-sound/qpwgraph/metadata.xml
```diff
@@ -0,0 +1,16 @@
...
+	<use>
+		<flag name="systray">Enables the system tray</flag>
+	</use>
+	<upstream>
```
NexAdn, line 11: Since you provide upstream information, please add a `<remote-id>` if possible. I don't know if there is a remote-id for third party GitLab instances, though. If not, just tell me and we can leave it out for now.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
NexAdn, line 4: We have EAPI=8 now. While EAPI=7 might have been the up-to-date version back when you submitted the PR, now EAPI=8 is the most recent version, so please bump the EAPI. The transition 7->8 is probably not that much work, since there were not major changes in the EAPI concerning “regular” ebuilds.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+inherit xdg cmake
+
+DESCRIPTION="Qt GUI to control the JACK Audio Connection Kit and ALSA sequencer connections"
+HOMEPAGE="https://gitlab.freedesktop.org/rncbc/qpwgraph"
+SRC_URI="https://gitlab.freedesktop.org/rncbc/qpwgraph/-/archive/v${PV}/${PN}-v${PV}.tar.gz"
```
NexAdn, line 10: Since you use the `${PN}-v${PV}` combination twice in the ebuild, you might wanna think about defining `MY_P="${PN}-v${PV}"` and using that instead to improve readability. That's just a suggestion, though.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="alsa debug systray"
+
+BDEPEND="virtual/pkgconfig"
```
NexAdn, line 17: In most ebuilds, `BDEPEND` is defined after `DEPEND` and `RDEPEND`. I suggest you stick with this convention.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
```diff
@@ -0,0 +1,39 @@
...
+DEPEND="
```
NexAdn, line 18: The libraries defined in this dependency list have subslots. It's usually a good idea to enable subslot rebuilds, i.e. depend on `cat/pkg:=` or `cat/pkg:slot=`. This ensures that the package is rebuilt if upstream changes the subslot (which usually corresponds to a change in the SONAME version, which requires a rebuild of reverse dependencies).

I heard that there are cases for Qt where subslot rebuilds are not necessary (depending on whether the private ABI is used by the software or not). However, I am not sure about this and if in doubt I suggest enabling the subslot rebuild to be safe.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+S="${WORKDIR}/${PN}-v${PV}"
+
+src_configure() {
+	local mycmakeargs=(
+		-DCONFIG_ALSA_MIDI=$(usex alsa 1 0)
+		-DCONFIG_DEBUG=$(usex debug 1 0)
+		-DCONFIG_SYSTEM_TRAY=$(usex systray 1 0)
+		-DCONFIG_WAYLAND=0
+		-DCONFIG_QT6=0
```
NexAdn, line 36: `usex` emits `true` or `false` and is thus a valid boolean value for CMake option. I suggest abbreviating these expressions to just `-DCONFIG...=$(usex USEFLAG)` and change `=0` to `=false` for the sake of uniformity.

media-sound/qpwgraph/qpwgraph-0.2.2.ebuild
NexAdn, line 18: You don't have `RDEPEND` defined. However, since you are linking against shared libraries, you will need to have at least those in `RDEPEND` as well. In this case, `RDEPEND="${DEPEND}"` is probably correct.

media-sound/qpwgraph/qpwgraph-9999.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	alsa? ( media-libs/alsa-lib )
+	systray? ( dev-qt/qtnetwork:5 )
+"
+# XXX OLD
+#dev-qt/qtx11extras:5
+#dbus? ( dev-qt/qtdbus:5 )
+
+#RDEPEND="${DEPEND}
+#	dev-qt/qtsvg:5
+#"
```
NexAdn, line 33: What are these comments for? Looks to me like they're some remnants of the development process or of outdated versions. They can probably be removed.

NexAdn (changes requested):
The ebuilds look quite good to me and they can probably be merged after the suggested changes have been applied. I have put comments only on the 0.2.2 ebuild. However, they apply to the 9999 ebuild as well, so please make sure to apply changes to both ebuilds.

Since we have updated our CI pipeline, please rebase your branch onto the current master so the new pipeline can run for your branch.

NexAdn:
One small thing I forgot to mention: please prefix your commit message with the catpkg of the package you are adding, i.e. something like `media-sound/qpwgraph: new package`. If you use pkgdev to commit your changes, it will write the correct commit message for you automatically in most cases.

mazes-80:
Can also have a look at guru overlay. I pushed this ebuild and also drumgizmo there.

Despite I thought audio-overlay to be a better place.

NexAdn:
> Can also have a look at guru overlay. I pushed this ebuild and also drumgizmo there.

Okay then. I guess as long as no one is interested in having the same ebuild in two places (and having to maintain the same ebuild twice), I'll close this PR for now. I don't think it's a good idea to have the same package in two ebuild repos.

PR #501, opened 2022-07-01
media-sound/bitwig-studio: Bump to 4.3

kaythomas0:


media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
```diff
@@ -0,0 +1,80 @@
...
+EAPI=7
+
+inherit desktop unpacker xdg
+
+DESCRIPTION="Multi-platform music-creation system for production, performance and DJing"
+HOMEPAGE="http://bitwig.com"
+SRC_URI="https://downloads.bitwig.com/stable/${PV}/${P}.deb"
+LICENSE="Bitwig"
+SLOT="0"
+KEYWORDS="~amd64"
```
NexAdn, line 13: Looking at the ebuild it seems like this is a precompiled package.
In this case, please use `KEYWORDS="-* ~amd64"` to indicate that this package won't work for arches other than amd64.

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
```diff
@@ -0,0 +1,80 @@
...
+
+IUSE="abi_x86_32 +jack cpu_flags_x86_sse4_1"
+REQUIRED_USE="cpu_flags_x86_sse4_1"
+
+DEPEND=""
+RDEPEND="${DEPEND}
+	app-arch/bzip2
+	dev-libs/expat
+	gnome-extra/zenity
+	jack? ( media-sound/jack2 )
```
NexAdn, line 24: Is `media-sound/jack2` really required or would `virtual/jack` do as well? Most applications are compatible with pipewire, so you'd usually want to depend on `virtual/jack` which can be provided by `media-sound/pipewire[jack-sdk]`.

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
```diff
@@ -0,0 +1,80 @@
...
+"
+
+QA_PREBUILT="
+	opt/bitwig-studio/.*
+"
+
+S=${WORKDIR}
+
+src_install() {
+	insinto /opt
```
NexAdn, line 60: insinto is not required if you don't use doins afterwards. If you only want to create a directory, dodir should be used instead.

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	cp -a opt/bitwig-studio "${ED}"/opt || die "cp failed"
+
+	dosym ../../opt/bitwig-studio/bitwig-studio /usr/bin/bitwig-studio
+
+	if ! use abi_x86_32; then
+		rm "${ED}/opt/bitwig-studio/bin/BitwigPluginHost-X86-SSE41" || die
+	fi
+
+	doicon -s scalable usr/share/icons/hicolor/scalable/apps/com.bitwig.BitwigStudio.svg
+	sed -i \
```
NexAdn, line 70: Such a sed call usually belongs to src_prepare(), not src_install(), since it modifies the source files.

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
```diff
@@ -0,0 +1,80 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
NexAdn, line 4: The most recent EAPI is EAPI=8. Please consider bumping the ebuild to the most recent EAPI.

NexAdn (changes requested):
Just found a few issues. Seems to me like they were already there in previous versions of the ebuild, but bumping to a new version is always a good opportunity to fix such issues :blush:

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
simonvanderveldt: FYI https://github.com/gentoo-audio/audio-overlay/pull/76#issuecomment-335279957

Could be this is fixed in Bitwig now (or that symbol is now also included/exposed in JACK), haven't checked

NexAdn: Hmm okay. That still doesn't rule out media-sound/pipewire[jack-sdk] as a potential candidate. Afaik its libjack.so.0 is fully ABI compatible with JACK 2's libjack.so.0. So maybe `|| ( media-sound/jack2 media-video/pipewire[jack-sdk] )` is the correct dependency here.

kaythomas0:
Closing this in favor of this PR: https://github.com/gentoo-audio/audio-overlay/pull/554

media-sound/bitwig-studio/bitwig-studio-4.3.ebuild
domichel: In the main tree, only jack-example-tools/jack-example-tools-4.ebuild depend on media-sound/jack2[libsamplerate], no package depend on media-sound/pipewire and around 70 or 80 software depend on virtual/jack. I would be very surprised is the correct depend for bitwig-studio is not virtual/jack. Also, to put a hard dependency on jack2 would break the virtual, which imply than if it is no imperative reason to break it, the ebuild should use virtual/jack.

Issue #503, opened 2022-08-15
Carla build fails.


apoorv569:
I am trying to install carla from this overlay but the build fails. I am pasting some output from the terminal,
```shell
text/String.cpp: In constructor 'water::String::String(const char*)':
text/String.cpp:297:81: error: 'numeric_limits' is not a member of 'std'
  297 |     CARLA_SAFE_ASSERT (t == nullptr || CharPointer_UTF8::isValidString (t, std::numeric_limits<int>::max()));
      |                                                                                 ^~~~~~~~~~~~~~
../../includes/CarlaDefines.h:168:54: note: in definition of macro 'CARLA_SAFE_ASSERT'
  168 | #define CARLA_SAFE_ASSERT(cond)               if (! (cond)) carla_safe_assert      (#cond, __FILE__, __LINE__);
      |                                                      ^~~~
text/String.cpp:297:96: error: expected primary-expression before 'int'
  297 |     CARLA_SAFE_ASSERT (t == nullptr || CharPointer_UTF8::isValidString (t, std::numeric_limits<int>::max()));
      |                                                                                                ^~~
../../includes/CarlaDefines.h:168:54: note: in definition of macro 'CARLA_SAFE_ASSERT'
  168 | #define CARLA_SAFE_ASSERT(cond)               if (! (cond)) carla_safe_assert      (#cond, __FILE__, __LINE__);
      |                                                      ^~~~
text/String.cpp: In constructor 'water::StringRef::StringRef(const char*)':
text/String.cpp:2048:67: error: 'numeric_limits' is not a member of 'std'
 2048 |     wassert (CharPointer_UTF8::isValidString (stringLiteral, std::numeric_limits<int>::max()));
      |                                                                   ^~~~~~~~~~~~~~
../../includes/CarlaDefines.h:168:54: note: in definition of macro 'CARLA_SAFE_ASSERT'
  168 | #define CARLA_SAFE_ASSERT(cond)               if (! (cond)) carla_safe_assert      (#cond, __FILE__, __LINE__);
      |                                                      ^~~~
text/String.cpp:2048:5: note: in expansion of macro 'wassert'
 2048 |     wassert (CharPointer_UTF8::isValidString (stringLiteral, std::numeric_limits<int>::max()));
      |     ^~~~~~~
text/String.cpp:2048:82: error: expected primary-expression before 'int'
[...]

Issue #476, opened 2021-08-20
New version v2.4.0 of carla available


gaobot:
A new version v2.4.0 of carla is available since 2021-08-19 21:30:02+00:00. For more details see https://github.com/falkTX/Carla/releases

Issue #457, opened 2021-04-14
New version v2.3.0 of carla available


gaobot:
A new version v2.3.0 of carla is available since 2021-04-13 18:25:05+00:00. For more details see https://github.com/falkTX/Carla/releases

0xC0ncord:
Looks like 2.3.0 also fixes a build issue on gcc-11

PR #504, opened 2022-08-20
media-sound/carla: Fix missing include in 2.2.0, add 2.5.2, drop 2.1.1, minor QA fixes

NexAdn:
A missing include in 2.2.0 caused build failures. This commit includes a
revbump for 2.2.0 with a patch adding this missing include.

Also there's a verbump to 2.5.2, which also bumps the EAPI version to 8.

Additionally, a few QA problems as indicated by pkgcheck have been fixed.

Closes: #503 
Closes: #457 
Closes: #476


NexAdn:
CI fails with 1, but says “No QA issues found”...

NexAdn:
@simonvanderveldt could you please have a look at this, since you're the maintainer of the package?

NexAdn:
ping

media-sound/carla/carla-2.5.2.ebuild
```diff
@@ -0,0 +1,102 @@
+# Copyright 1999-2022 Gentoo Authors
```
simonvanderveldt, line 1: Can this be kept a symlink to the 9999 ebuild?

media-sound/carla/carla-9999-r1.ebuild
```diff
@@ -23,7 +22,7 @@ fi
 LICENSE="GPL-2 LGPL-3"
 SLOT="0"
 
-IUSE="alsa gtk gtk2 opengl osc -pulseaudio rdf sf2 sndfile X"
+IUSE="alsa gtk gtk2 opengl osc pulseaudio rdf sf2 sndfile X"
```
simonvanderveldt, line 25: I guess this is removed because of some `pkgcheck` output? Are there any docs that explain why it's flagging this?

media-sound/carla/Manifest
```diff
@@ -1,2 +1,3 @@
 DIST carla-2.1.1.tar.gz 9177989 BLAKE2B 87f5fe97e3d3e4d7f25da4b14464069b6cd1c83405c458b9d80add3e54d61d59763f074a3a85c9c88efdcd8f5790303992cc3233b767189cd4f50c60cbeda892 SHA512 e8da99d7bddff1415dd54ac3e1abe90fbccb42877d2daec0a60ad046e6d15531ebf8328c4d44c06cb14a7a5308d43be62d88eed6268a55dfbfae0f620a5f04dc
 DIST carla-2.2.0.tar.gz 9844747 BLAKE2B 1fd778a1c2904fe982c3aec32e9bb39fff679bea0fe8d1fe41d10bbc5347dd82a6f3dcd33d19aee8bc3b123e387a8b7655c9735fca784d642596f0df9085e38e SHA512 afb2beab979f801817a7fd7663bfdc90800db021679243faa5acb15de60f6dc281032e05c51a8625124c69e7759099fc682678eb98c08507f952763e665ac9c4
```
simonvanderveldt, line 2: We can just drop the old versions, generally we keep/have kept live, latest and latest-1 (where -1 is mostly the major or minor version)

media-sound/carla/carla-2.5.2.ebuild
NexAdn: It might. However, this is for one not what ::gentoo does (and I assume that they have enough experience to have at least tried or discussed using this method) and on the other hand I personally had a few problems with it:

* Although a non-issue in this overlay, it makes stabilization and in general setting KEYWORDS differently for different versions impossible
* If need for patching certain versions arises (e.g. to backport a fix), a symlink can not be used unless the patch is required for _all_ versions (including -9999)
* I would try to stick to the workflow used in ::gentoo just for the sake of not having to think about whether I commit to ::gentoo or ::audio-overlay.
* When applying changes to the -9999 ebuild, one must never forget to do check if this requires a revbump for the versioned ebuilds. This is easier to forget if one only has to edit -9999 instead of each versioned ebuild.
* This approach only works if the dependency and IUSE specifications are _exactly_ the same for each version (which breaks as soon as upstream introduces a new or removes an old dependency)
* It doesn't bring a noticeable reduction in size on users' devices (especially with Portage now keeping the git history for auto-synced repos to a minimum by using shallow clones and the like). At least not compared to the size of the ::gentoo repo (which we depend on).

I know that most of these points will probably be never or almost never a issue in reality. However, I've already had to deal with backporting fixes to certain versions or introducing new IUSE in newer versions that weren't backwards compatible to older versions. I've used the symlink-based workflow before and it brought me more problems than it made things simpler. And I'd rather have no symlinks in the off-chance we might need it in the future (which is becoming more likely with languages like Go and Rust, which pin their dependencies, growing in popularity or languages which are known for having many dependencies per package in general). We could switch to a copy-based workflow for single packages if needed. However, for the sake of uniformity I'd rather have a single policy for the whole overlay instead of some packages violating a common workflow.

Still, it is your overlay, so if you insist I'll revert to the symlink based workflow. Just be aware of the possible implications.

media-sound/carla/carla-9999-r1.ebuild
NexAdn: This is from pkgcheck output. Unfortunately, I couldn't find detailed documentation on this issue, yet.
However, it is one of the red checks (aka. “critical”) and thus one which shouldn't be ignored if we don't have a particular reason to do so.

My thinking why this IUSE configuration was flagged is probably because of how IUSE defaults are processed.
According to the [devmanual](https://devmanual.gentoo.org/general-concepts/use-flags/index.html#iuse-defaults) adding a `-` before a USE flag is essential a NOOP as the setting is always overriden by the profile make.defaults, package.use or make.conf if the respective USE flag is set. If it is not set there, the USE flag is disabled anyway, so it really shouldn't matter if `-` is specified in IUSE or not. For the sake of not confusing developers (and preventing issues in case Portage decides how to interpret such a `-` in the future), omitting the `-` is a sensible choice.

media-sound/carla/Manifest
NexAdn: Well, that wasn't the initial purpose of this PR, but I'll do that. I don't see the need for keeping more than two versions around, anyway.

NexAdn: dropped 2.1.1

media-sound/carla/carla-2.5.2.ebuild
simonvanderveldt: FYI the reason for using symlinks is that otherwise the git log/diff is useless. It's one of the IMHO few/main issues with how Gentoo/Portage/PMS handles packages. So I would always use symlinks to ensure everyone has an easy way to check what has changed, be it in a PR or when just looking through the git log.
Most of the points you mentioned you can do with symlinks, but there's of course a balance to be struck, you don't want 10 conditionals on versions in your ebuild, so sometimes a versioned one makes more sense (I've pretty much only done this with patches that are no longer needed for newer versions or the live ebuild. I've ran into few if any (can't remember any) issues using this approach).

audiodef (approved):

audiodef (approved):

audiodef (approved):

audiodef (approved):

audiodef (approved):

media-sound/carla/carla-2.5.2.ebuild
NexAdn: This is a sensible explanation. I've kept 2.2.0 as a copy, though, since it requires applying a patch to build. 2.5.2, however, is now a symlink to 9999, conforming to your workflow.

media-sound/carla/carla-9999-r1.ebuild
NexAdn: Looks like there is no further opposition to my explanation, so I consider this resolved.

NexAdn:
Review can wait. Looks to me like we have pugl as a new dependency for Carla.
I guess this is pulled in only if no other graphics backend is enabled since I didn't get these compiler errors from the CI pipeline, although I don't have pugl installed.

NexAdn:
The build also fails locally if `USE=-X` is set. Maybe we have to set new options for the Makefile to make the build work with `USE=-X`.

NexAdn:
I'm thinking about merging this despite the pipeline failure. Emerging this ebuild only fails if `USE=-X` is set and I suspect most (if not all) users have `USE=X` set and I consider it more critical to fix the global scope failure (which prevents the package from being installed at all) than to fix the build failure which occurs only for certain USE flag combinations.

WDYT?

audiodef (approved):

PR #505, opened 2022-09-09
media-libs/pyliblo: Add tests, PEP517 and recent python versions

NexAdn:
These commits go alongside each other. Enabling tests is required for testing the python bumps.
Only using PEP517 is a minor quality improvement and independent of the rest of the changeset.

This PR is a requirement for building pyliblo on recent Gentoo systems without having to enable py3.9 support.

### DIFF
--- a/media-libs/pyliblo/pyliblo-0.10.0-r2.ebuild
+++ b/media-libs/pyliblo/pyliblo-0.10.0-r2.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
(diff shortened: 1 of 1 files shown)


NexAdn:
CI pipeline complains about unset `GITHUB_TOKEN` variable, but the actual check runs through without problems.

NexAdn:
Rebased onto current master.

NexAdn:
This is a rather small change and shouldn't cause any problems. Merging without review.

PR #510, opened 2022-12-06
profiles: Drop pyliblo unmask

NexAdn:
The unmask was added to circumvent problems during the migration of media-libs/pyliblo from ::gentoo to ::audio-overlay. Since it is now obsolete, this commit drops the unmask.


simonvanderveldt:
Correct, the mask was dropped in https://github.com/gentoo/gentoo/commit/7e2ab3e3c33cae1b933b5cee2cc47bdac964fcbc so this is indeed no longer necessary. Thanks!

simonvanderveldt (approved):

PR #511, opened 2022-12-09
Misc QA fixes

NexAdn:
This PR contains a set of fixes as indicated by pkgcheck as well as a few `PYTHON_COMPAT` updates (dropping versions no longer supported in ::gentoo).
Overall these changes should not affect the build process itself.

### DIFF
--- a/media-libs/libmonome/libmonome-9999.ebuild
+++ b/media-libs/libmonome/libmonome-9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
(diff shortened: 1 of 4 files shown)


NexAdn:
As @audiodef pointed out in IRC, the added `${PYTHON_DEPS}` is more or less in conflict with #518. While this is technically true, these are not actually conflicts sematically. `${PYTHON_DEPS}` is a dependency for the build system an thus belongs to `BDEPEND`, which is not available in EAPI=6 and thus can't be used in this PR.

When merging both PRs, it's probably best to just drop the commits for the `${PYTHON_DEPS}` in this PR and let #518 handle adding the deps in the right place.

NexAdn:
Branch rebased onto current master. Conflicts with #518 are now resolved.

audiodef (approved):

audiodef (approved):

audiodef:
Fails check, but hopefully with pkgcheck replacing repoman we can clear this up.

NexAdn:
The pipeline just complains about `Please set the GITHUB_TOKEN environment variable` :upside_down_face: The actual build of the packages seems to have worked, so I'll merge despite the pipeline complaining.

Issue #515, opened 2022-12-26
Migrate consumers of waf-utils.eclass to EAPI > 6


NexAdn:
Today there was a patchset on gentoo-dev for dropping EAPI=6 support in waf-utils.eclass. Since we have a few packages which require waf-utils.eclass and are still on EAPI=6, we must expect ebuild failures in global scope to occur (and thus our ebuilds to break) soon.

Thus, we need to check the current list of EAPI=6 ebuilds if they can be bumped to EAPI=7 or EAPI=8:

```
$ grep -nR EAPI=6
media-libs/libmonome/libmonome-1.4.3.ebuild:4:EAPI=6
media-libs/libmonome/libmonome-9999.ebuild:4:EAPI=6
media-plugins/bitrot/bitrot-9999.ebuild:4:EAPI=6
media-plugins/deteriorate-lv2/deteriorate-lv2-1.0.7-r2.ebuild:4:EAPI=6
media-plugins/deteriorate-lv2/deteriorate-lv2-9999.ebuild:4:EAPI=6
media-sound/bitwig-studio/bitwig-studio-1.3.16.ebuild:8:EAPI=6
media-sound/bitwig-studio/bitwig-studio-2.5.1.ebuild:8:EAPI=6
media-sound/bitwig-studio/bitwig-studio-3.0.3.ebuild:8:EAPI=6
media-sound/bitwig-studio/bitwig-studio-3.2.8.ebuild:4:EAPI=6
media-sound/bitwig-studio/bitwig-studio-3.1.3.ebuild:4:EAPI=6
media-sound/cadence/cadence-9999-r7.ebuild:4:EAPI=6
media-sound/sequencer64/sequencer64-0.96.7.ebuild:4:EAPI=6
media-sound/sequencer64/sequencer64-0.96.8.ebuild:4:EAPI=6
media-sound/sequencer64/sequencer64-9999.ebuild:4:EAPI=6
media-sound/serialosc/serialosc-1.4.1.ebuild:4:EAPI=6
media-sound/serialosc/serialosc-9999.ebuild:4:EAPI=6
x11-libs/ntk/ntk-1.3.1001.ebuild:4:EAPI=6
x11-libs/ntk/ntk-9999.ebuild:4:EAPI=6
```

NexAdn:
Found these waf-utils consumers with EAPI=6:

* libmonome-1.4.3.ebuild
* bitrot-9999.ebuild
[...]

PR #518, opened 2023-01-03
*/*: Bump consumers of waf-utils.eclass to EAPI 8

NexAdn:
waf-utils.eclass no longer supports EAPI=6. This PR bumps all consumers which still used EAPI=6 to use the most recent EAPI=8.

Closes: https://github.com/gentoo-audio/audio-overlay/issues/515


audiodef (approved):
Agreed.

NexAdn:
pkgcheck run locally. These ebuilds no longer cause global scope failures. Thus merging despite pipeline failure.

PR #519, opened 2023-01-03
metadata: Deprecate EAPI 6

NexAdn:
EAPI 6 is already very old and almost gone from ::gentoo. It is likely that it will be banned there sooner or later. It should at least be deprecated here and banned after migrating all ebuilds to newer EAPIs.


audiodef (approved):
I agree. As of January 2023, ebuilds still using EAPI 6 will fail.

simonvanderveldt (approved):

PR #520, opened 2023-02-06
ci: replace repoman with pkgcheck

NexAdn:
This PR is a partial rework of #512, introducing pkgcheck as a replacement for repoman.
This PR, compared to its predecessor, is kept as similar to the repoman workflow as possible
as there was no consensus about the major rework in #512, yet.

The manifest checks from #512 are not introduced in this PR as they require the [fem-overlay-ci-tools](https://gitlab.fem-net.de/gentoo/fem-overlay-ci-tools/) which were provided in the [fem-overlay-ci-image](https://gitlab.fem-net.de/gentoo/fem-overlay-ci-image/).
Introducing these checks later is not impossible,
but I'd like to have the discussion in #512 resolved before I put unnecessary work into this.


audiodef (approved):

audiodef (commented):
Yes! I'm not a fan of CI leaving comments anyway.

.circleci/config.yml
```diff
@@ -31,28 +31,32 @@ jobs:
...
     resource_class: large
     steps:
       - checkout
       - restore_cache:
-          key: portage-pkgdir-repoman
-      - run: ./tests/repoman-new-or-changed-ebuilds.sh
+          key: portage-pkgdir-pkgcheck
+      - run: |
+          git fetch
+          ./tests/pkgcheck.sh --commits origin/master..HEAD
```
simonvanderveldt, line 56: Nice that `pkgcheck` support determining the new/changed ebuilds out of the box :)

simonvanderveldt (commented):
Thanks for fixing this! :+1: 

Does `pkgcheck` fail on ebuilds that don't start with
```
# Copyright <year>-<year> Gentoo Authors
# Distributed under the terms of the GNU General Public License v2
```
?

.circleci/config.yml
NexAdn: > Thanks for fixing this! +1
> 
> Does `pkgcheck` fail on ebuilds that don't start with
> 
> ```
> # Copyright <year>-<year> Gentoo Authors
> # Distributed under the terms of the GNU General Public License v2
> ```
> 
> ?

From pkgcheck(1) there seem to be a few report types for this:

* EclassIncorrectCopyright
* EclassInvalidCopyright
* EclassNonGentooAuthorsCopyright
* EclassOldGentooCopyright
* IncorrectCopyright
* InvalidCopyright
* NonGentooAuthorsCopyright
* OldGentooCopyright

(I give no guarantee that this is the complete list)

However, at least some (if not all) of these checks are enabled per default only for ::gentoo. We might still be able to enable some of these for our repo in metadata/pkgcheck.conf and even make the ones we'd like to enforce critical failures (like I did for BadFilename, MissingUri and MissingLicenseRestricts). This is mostly just a matter of configuring pkgcheck as we like.

However, ensuring those copyright headers exist is a job which can be done manually with ease, so if we don't manage to automatically check this, it shouldn't be hard to manually ensure the headers exist. And as a last resort, grepping ebuilds in the pipeline for these headers is not very hard either.

simonvanderveldt: OK, that's nice. I don't have any experience with pkgcheck yet, that's why I'm asking. repoman didn't allow for configuring stuff like this resulting in overlays being "required" to write the same headers as are required in `::gentoo` or getting a warning that you had to ignore, which is annoying.
Good to see that it looks like pkgcheck is an improvement in this regard.
I don't think we need to spend time/bikeshed now on these configs btw, was just interested if this known issue with repoman was also an issue with pkgcheck :) thanks for the info!

PR #523, opened 2023-02-08
Delete pure-data-9999.ebuild

audiodef:

PR #524, opened 2023-02-08
Add files via upload

audiodef:
media-sound/pure-data: Rm 9999 version, rm symlinks to 9999, add 53.1


NexAdn:
I see a few issues with this PR:

* The summary doesn't really tell me what it does. That makes it hard to browse through the PR list.
* Missing description of anything. This PR (or rather the commit) lacks a description of why some of these changes happen. In particular (looking at #523 as well) there is no explanation for why the live ebuild is deleted and why Simon is removed as a maintainer (which is something Simon should either do himself or ACK on, so I won't merge this in either case without Simon's approval).
* Your commit does two things at once, namely removing a maintainer and adding a new version. The changes in the Manifest don't match the addition of the new version, since they also drop the old entries as well.

My suggestion for a fix:

* Start with the master branch.
* Checkout a new branch: `git checkout -b my-new-branch-name`
* Add the ebuild for the new version. Following Simon's workflow, I suggest you stick with symlinks unless you have a good reason not to do so.
* Regenerate the manifest using `pkgdev manifest`.
* Test the ebuild.
* Commit your changes using `pkgdev commit`. This will run QA checks and write a proper commit description if everything goes well.
* Remove the old version, regenerate the manifest using `pkgdev manifest` and commit the changes using `pkgdev commit`.
* Push your changes to your personal fork.
* Create a new PR.

You might be able to switch this PR to a fresh branch, but I am not sure. Creating a new PR is probably easier. Depending on how you'd like to do it, keep this PR open or close it as you think is right.

audiodef:
Real quick, will get to more later, I didn't remove anyone. The interface told me there was no maintainer so that was just me trying to be helpful. You'll have to blame the interface for that.

PR #539, opened 2023-03-16
tests: do not install unneeded pip dependency

NexAdn:
This old tool is no longer in use and causes problems in the CI pipeline.


NexAdn:
Pipeline seems to work and this is a small but critical fix. Merging without review.

Issue #528, opened 2023-02-24
media-sound/carla: Fails to build with USE=-X -gtk -gtk2


NexAdn:
See also https://circleci.com/gh/gentoo-audio/audio-overlay/5366.
Carla doesn't like it when we disable all display backends.

PR #543, opened 2023-05-26
media-sound/carla: enable py3.11

vi-improved:
Enables py3.11 and cleans up removed python targets.

### DIFF
--- a/media-sound/carla/carla-9999-r1.ebuild
+++ b/media-sound/carla/carla-9999-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6,7,8,9,10} )
+PYTHON_COMPAT=( python3_{10,11} )
 inherit python-single-r1 xdg-utils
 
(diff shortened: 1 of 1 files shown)


vi-improved:
> Did you test the software with py3.11? We've had experience with build failures for a few packages in the overlay while testing py3.11 compatibility, so I'd like to be sure that we don't break carla by enabling py3.11. A quick “yes” is enough.
> 
> The PR in general looks fine, but the pipeline failure does not look good. Seems like the package is unable to install the files correctly.

I have tested it with py3.11, and it works fine. No idea why the pipeline failure is happening. I'll check it out.

vi-improved:
I can't reproduce the CI failure on my machine, even with the same USE flags.

NexAdn:
Tested your PR locally. I can't reproduce the build failure either. This might be related to https://github.com/gentoo-audio/audio-overlay/issues/528, so I'll merge this despite the pipeline failure. I really don't think that this has anything to do with the PYTHON_COMPAT bump and I suspect it won't even affect any user.

Thanks for your contribution!

PR #550, opened 2023-08-15
media-plugins/distrho-ports: add version 2021-03-15

### LINKED ISSUE #455: New version 2021-03-15 of distrho-ports available
A new version 2021-03-15 of distrho-ports is available since 2021-03-15 19:51:21+00:00. For more details see https://github.com/DISTRHO/DISTRHO-Ports/releases

nicoroeser:
Not the most sophisticated piece of work ever accomplished, but works for me.

Fixes #455.


NexAdn:
Seems like the CI pipeline is unable to build the package due to missing dependencies (fftw3f):

https://app.circleci.com/pipelines/github/gentoo-audio/audio-overlay/2005/workflows/2eb55cf6-0305-46e0-bde4-ba46f1d70e1b/jobs/5914

Could you please have a look at the dependencies?

nicoroeser:
I have added a commit which takes care of the additional dependency.

Is this conditional syntax acceptable in live ebuild files?

If not, my plan would be to _copy_ the ebuild to 20210115 (which does not yet depend on FFTW) and make the change to the live ebuild (unconditionally) afterwards. Does that sound better?

NexAdn:
I'm not really a fan of symlinked ebuilds and putting if blocks in the live ebuild for every extra detail just makes the ebuild overly complicated. Hence, I'd prefer if you don't use symlinks and instead copy the files and add the dependency only in the live ebuild.

nicoroeser:
I agree. This pull request now removes symlinks and provides three separate ebuilds: the last two released versions of DISTRHO-ports and the live ebuild.

It also adds a patch for version 2021-01-15 because that version did not build with GCC 12 or 13 (have tried GCC versions 11, 12, and 13). If you decide to rather drop the ebuild for DISTRHO-ports version 2021-01-15, that’s fine for me, too. In this case, the patch can be removed as well as it is not required for later versions.

NexAdn:
Nice job! Unfortunately, the CI pipeline fails:

> Did not find CMake 'cmake'
> Found CMake: NO
> Run-time dependency x11 found: NO (tried pkgconfig)
> 
> meson.build:91:8: ERROR: Dependency "x11" not found, tried pkgconfig

Again, this looks like something is wrong with the dependencies. Maybe there's some new configurable options which can be sensibly translated to USE flags?

You might wanna try using dev-util/ebuildtester to test the ebuild in a clean environment (only if you have Docker, though).

So, in general I'm okay with this PR. It's just the missing dependencies or configure options that need fixing.

nicoroeser:
Thanks for the hint about `dev-util/ebuildtester`!

I haven’t been able to reproduce the issue. When trying the ebuilds with `ebuildtester`, they work in my setup. I have also tried the CI script directly:

```
./tests/emerge-ebuild.sh =media-plugins/distrho-ports-20210115
```

This works as well. The (slightly extended) part of the build log that corresponds to the part which failed on CircleCI is:

```
Found pkg-config: /usr/bin/x86_64-pc-linux-gnu-pkg-config (1.8.1)
Run-time dependency alsa found: YES 1.2.9
Run-time dependency freetype2 found: YES 25.0.19
Run-time dependency x11 found: YES 1.8.6
Run-time dependency xext found: YES 1.3.5
Library dl found: YES
Run-time dependency gl found: YES 1.2
Build targets in project: 139

```

I have also tried `pkg-config --libs x11` in this container (while the build was running), again without errors.

That means either something is very wrong on my end or in the CircleCI pipeline (or both, or something completely different is going on).

As far as I understand it, a cache `portage-pkgdir-emerge` is used on CircleCI for the binary packages. Is it possible that the binary package `x11-libs/libX11-1.8.6-1` is lacking `/usr/lib64/pkgconfig/x11.pc`? Or that `dev-util/pkgconf` is not OK? Maybe a bit far-fetched, but I’m out of other ideas right now.

NexAdn:
Sorry, but I'm currently a very busy. I can't promise that I'll have a look at this until october. I hope that's okay!

media-plugins/distrho-ports/distrho-ports-20210315.ebuild
```diff
@@ -0,0 +1,49 @@
+# Copyright 1999-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
NexAdn, line 4: We are on EAPI=8 now. Please consider bumping the EAPI.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -1,22 +1,14 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
```
NexAdn, line 4: As before, please consider an EAPI bump.

media-plugins/distrho-ports/distrho-ports-99999999.ebuild
```diff
@@ -1,22 +1,14 @@
...
-if [[ ${PV} == *9999 ]]; then
-	inherit git-r3
-	KEYWORDS=""
-else
-	MY_PN="${PV:0:4}-${PV:4:2}-${PV:6:2}"
-	SRC_URI="https://github.com/DISTRHO/DISTRHO-Ports/archive/${MY_PN}.tar.gz -> ${P}.tar.gz"
-	KEYWORDS="~amd64"
-	S="${WORKDIR}/DISTRHO-Ports-${MY_PN}"
-fi
+KEYWORDS=""
```
NexAdn, line 11: Is there a reason to remove the `if` block? Keeping the `if` block would allow us to only maintain the live ebuild and copy it over to new versioned ebuilds upon new releases.

NexAdn (changes requested):
Just a few minor questions, but looks fine overall.

PR #555, opened 2023-12-15
media-sound/pure-data: bump to 0.54.1 and add support for double-precision

kennypm:


media-sound/pure-data/pure-data-9999.ebuild
```diff
@@ -40,6 +40,7 @@ src_prepare() {
 
 # Disable portaudio and portmidi because otherwise Pd's local sources get installed
 src_configure() {
+	use double-precision && append-cppflags "-DPD_FLOATSIZE=64"
```
NexAdn, line 43: Are there any benefits or downsides from using nor not using the 64 bit downsides that would legitimize providing a USE flag for this?

NexAdn: floatsize*

kennypm: Examples of use cases requiring doubles include very low frequency oscillators and indexing large arrays, especially with interpolation. More details from someone who did a lot of work expanding Pd to double precision can be found [here](https://www.katjaas.nl/doubleprecision/doubleprecision.html).

Backwards compatibility is a high priority for Pd so I believe that's a big reason why they haven't switched to doubles by default. Theoretically there could be performance overhead as well but it hasn't been significant enough to be noticeable for me.

`equery hasuse double-precision` returns the following for me:
```
[IP-] [  ] media-libs/aubio-0.4.9-r2:0/5
[IP-] [  ] sci-libs/metis-5.2.1-r2:0
```
So there is some precedent for this. I just took the description of the flag from `sci-libs/metis`.

The configure script now has an optional argument `--with-floatsize=<SIZE>` that does the same thing if you'd prefer not using `flag-o-matic.eclass`, I just wrote this ebuild before they included that option and didn't think to change it.

NexAdn: okay then

NexAdn (approved):

Issue #568, opened 2025-05-17
Removal: x11-libs/ntk and friends


NexAdn:
These packages are dead upstream for years now. We can expect the situation won't get better and must consider Non DAW dead. Removal starts now.

PR #575, opened 2026-04-04
media-sound/luppp: treeclean

BakedSnake:
Depends on removed package _ntk_. 

- #570 
- #568 

The project's _github_ has other _ntk_ sources that are also not available anymore.


BakedSnake:
Converting to draft, since it may still be possible to bump ntk

